Issue #5249, opened 2014-08-05
Nokia Lumia 928 + IE Mobile 11


tzi:
Hi Piwik team!

The IE Mobile team [has announced a new UA](http://blogs.msdn.com/b/ie/archive/2014/07/31/the-mobile-web-should-just-work-for-everyone.aspx) that simulate an iPhone:

    Mozilla/5.0 (Mobile; Windows Phone 8.1; Android 4.0; ARM; Trident/7.0; Touch; rv:11.0; IEMobile/11.0; NOKIA; Lumia 928) like iPhone OS 7_0_3 Mac OS X AppleWebKit/537 (KHTML, like Gecko) Mobile Safari/537

So I tried it with our new device-detector:

    os:
      name: Android
      short_name: AND
      version: 4.0
    client:
      type: browser
      name: IE Mobile
      short_name: IM
      version: 11.0
    device:
      type: smartphone
      brand: NK
      model: Lumia 928
    os_family: Android
    browser_family: Internet Explorer

The problem is actually not about the browser, but the OS that should be an Windows Phone.

Cheers,  
Thomas.

nickurt:
Isn't it possible to do something with 'a priority'?, 

When moving the 'Windows Mobile/Phone' regexes (oss.yml) just above the Android regexes it will return the correct information, with no tests failing ...

```
array(5) {
  ["type"]=>
  string(7) "browser"
  ["name"]=>
  string(9) "IE Mobile"
  ["short_name"]=>
  string(2) "IM"
  ["version"]=>
  string(4) "11.0"
  ["engine"]=>
  string(7) "Trident"
}
array(3) {
  ["name"]=>
  string(13) "Windows Phone"
  ["short_name"]=>
  string(3) "WPH"
  ["version"]=>
  string(3) "8.1"
}
int(1)
string(2) "NK"
string(9) "Lumia 928"
```

- - - 
Just added this as new test

```
-
  user_agent: Mozilla/5.0 (Mobile; Windows Phone 8.1; Android 4.0; ARM; Trident/7.0; Touch; rv:11.0; IEMobile/11.0; NOKIA; Lumia 928) like iPhone OS 7_0_3 Mac OS X AppleWebKit/537 (KHTML, like Gecko) Mobile Safari/537
  os:
    name: Windows Phone
    short_name: WPH
    version: 8.1
```

I can do a pull requests if needed...

PR #5294, opened 2015-01-01
Support for the new UA of Windows Phone 8.1 to simulate an iPhone

nickurt:
See #5249


sgiehl:
Thanks for the PR. I will merge it. Hope there won't be too many of those ugly UAs, as we might run into trouble trying to adjust detection for every one.

PR #5295, opened 2015-01-01
Ignore memcache tests when the extension is not loaded ..

nickurt:


sgiehl:
I'll merge it for the next release. But in 3.0 the caching stuff will be replaced with doctrine/cache or something similar.

PR #5297, opened 2015-01-09
Update DeviceDetector VERSION to 2.8

xavren:


sgiehl:
already working on 3.0. So updating to 2.8 on master is too late

PR #5305, opened 2015-01-26
Add detection of WeChat and Weibo apps

mexon:
Hi,

I want to use piwik to replace Baidu analytics.  But it's not much use for a predominantly Chinese website because it doesn't detect WeChat.  These days we're finding our website gets over 50% of incoming hits from WeChat.

I don't have much idea what I'm doing here, or what schema you have in mind for short_name.  But this seems like a start.


sgiehl:
Thanks for you PR. Adding detection for WeChat and Weibo apps is nice. There is no need to introduce short codes for mobile apps. They are kind of legacy for browsers and maybe os aswell. Please revert your changes to ```Parser/Client/MobileApp.php``` and remove the short codes from the tests, where they don't apply.

mexon:
OK, I've taken away the short codes.  I'm using Piwik 2.10.0 and I was hoping to have nice logos for WeChat and Weibo, which seems to depend on short codes.  But no problem, I can hack up my own Piwik until these changes make it back downstream.

sgiehl:
Sad to say that, but piwik currently is not able to handle detected mobile apps. It's only showing the browsers. There is an open ticket for that. But it will take a while to implement it.

PR #5308, opened 2015-01-29
galaxy note detections

helbling:
see commit messages


sgiehl:
THX for the PR. Could you please fix the existing tests and maybe add some new tests for Galaxy Note 4 devices if you have some example useragents?

helbling:
done, fixed old tests and added tests for the new UA strings

PR #5312, opened 2015-02-10
Add QQ browser

mexon:
QQ Browser is a somewhat popular browser in China.

### DIFF
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -115,6 +115,7 @@ class Browser extends ClientParserAbstract
         'PW' => 'Palm WebPro',
         'PX' => 'Phoenix',
         'PO' => 'Polaris',
+        'QQ' => 'QQ Browser',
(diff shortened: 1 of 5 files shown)


sgiehl:
Thanks for the PR.
Please add the browser name to the shortcode list in https://github.com/piwik/device-detector/blob/master/Parser/Client/Browser.php#L119
And please add a test at least in https://github.com/piwik/device-detector/blob/master/Tests/Parser/Client/fixtures/browser.yml

mexon:
Ah, turns out adding a test was a good move since the code I had before didn't actually work :-(

sgiehl:
Ok. Looks better now ;)
Another small question? Are you sure QQBrowser uses Gecko engine? The UA examples you added as test cases look more like it uses WebKit. If you are not sure about the enige, just leave it away. DeviceDetector will then try to autodetect (what will result in WebKit in this cases)

PR #5314, opened 2015-02-12
All iPhone submodels (S, C) are capitalized. Update Airness device regex.

peteygao:
I updated the model name for iPhone 5C/5S since the other iPhone versions are already capitalized (`iPhone 4S`, `iPhone 3GS`, etc).

Also I updated the Airness device regex as the headers look something like this:
`AIRNESS-AIR99/REV 2.2.1/Teleca Q03B1`

### DIFF
--- a/regexes/device/mobiles.yml
+++ b/regexes/device/mobiles.yml
@@ -224,10 +224,10 @@ Apple:
       model: 'iPhone 5'
       device: 'smartphone'
     - regex: '(?:Apple-)?iPhone5[C,][34]'
-      model: 'iPhone 5c'
+      model: 'iPhone 5C'
       device: 'smartphone'
     - regex: '(?:Apple-)?iPhone6[C,][12]'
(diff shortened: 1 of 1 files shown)


sgiehl:
Would you please update the tests aswell, and maybe also add a test case for an airness device if you have one.

peteygao:
I don't want to sound like I'm copping out, but I'm not a PHP developer. I'm contributing to a [Ruby port of this project](https://github.com/podigee/device_detector/pull/2) and I'm going through the Regexes and simply posting upstream the fixes/changes I'm making downstream.

I don't know how I'd run PHP tests even if I do make modifications to them. I'm happy to learn if there's a simple way to set up PHP and the required testing frameworks.

I also do not have an Airness device, however I did find some online examples for Airness device user agents (which is where I got that UA from).

sgiehl:
Ok. No problem. I'll fix the tests after merging

PR #5322, opened 2015-03-17
Move NetCast regex from mobiles to televisions

iagooar:
I have discovered, that NetCast detection is partially done through the `mobiles` regexes, instead of `televisions`. I have merged both regexes in `televisions.yml`


yagooar:
BTW, shouldn't the device name actually be `LG`?

sgiehl:
Sorry, that can't be simply moved. The televisions.yml is only parsed for devices having "HbbTV/X.X" within the useragent. That is done to speed up the detection as otherwise all detections would be done.

PR #5335, opened 2015-04-18
Merge pull request #1 from piwik/master

everleek:
合并官方


sgiehl:
Not sure what you aimed with that pull request, but there are no changes in it...

PR #5337, opened 2015-04-20
Better detection of Playstation 3

matt-kv:


sgiehl:
The mentioned user agent should already be detected correctly. All regexes are used case insensitive, so there is no need to have that in different cases

matt-kv:
Ok, I hadn't seen that, sorry.

PR #5339, opened 2015-04-21
addition of the Qooq tablet

matt-kv:
I found ua's for this tablet in my logs. Therefore I'm submitting a regex for detection and a test ua.


sgiehl:
Please also add a short code for that brand in https://github.com/piwik/device-detector/blob/master/Parser/Device/DeviceParserAbstract.php#L259
Afterwards you also need to put that in the test fixture

PR #5345, opened 2015-05-14
Add Kodi (XBMC after version 14)

zenermerps:
Added the Kodi mediaplayer to the list, which is XBMC that was renamed for version 14.0.

### DIFF
--- a/regexes/client/mediaplayers.yml
+++ b/regexes/client/mediaplayers.yml
@@ -65,6 +65,10 @@
   name: 'XBMC'
   version: '$1'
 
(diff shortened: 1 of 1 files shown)


sgiehl:
Would you mind adding a test or providing an example useragent to make one?

ToddShepard:
Of course, here you go: 

```
Kodi/14.0 (Macintosh; Intel Mac OS X 10_10_3) App_Bitness/64 Version/14.0-Git:2014-12-23-ad747d9-dirty
```

PR #5347, opened 2015-05-17
Improved dectection of Wiko smartphones

matt-kv:
I modified the regexes somewhat, added some models, and substituted a generic naming rule instead of the model-by-model rules.
Unfortunately that means that the resulting model names are often in CAPITALS, so the tests have to be changed accordingly.
If this not ok, specific rules for each model have to be reintroduced, and tests have to be written for each new model.


sgiehl:
You already mentioned the big disadvantage of combining the rules. The model names are not unique anymore. So there might be Highway, HIGHWAY, HighWay or whatever the useragent contains. That means each tool using DeviceDetector needs to handle that correctly.
I know that affects the most general detection rules, but I'm currently not sure how to handle that correctly. Specifying each device name (like for wiko) is one option, but maybe there would be a better one...

matt-kv:
The only thing I can think of is trying to manipulate upper/lowercase in the '$1' in order to normalize them.
In the case of Wiko models, I tried using `ucwords(strtolower('$1'))` but the whole is considered as a string so it doesn't work...

matt-kv:
I don't know php, so I don't know whether this idea (manipulating the contents of the '$1' from within the yml file) is even doable... It's probably better to revert to specific regexes for each model.

sgiehl:
Yes. guess it's better to revert to specific regexes for now. I need to think about a proper solution before merging regexes...

matt-kv:
Ok, I reverted to specific cases for each model and modified the tests accordingly.

matt-kv:
Sorry if it's not totally clear, the changes have been made within this pull request...

sgiehl:
Sorry for the delay. Lgtm now. Merging...

PR #5351, opened 2015-06-03
Added detection for Firefox iOS

tzi:
Hi!

Mozilla chose the UA for Firefox iOS:
https://bugzilla.mozilla.org/show_bug.cgi?id=1147658

Warning, this Firefox is running over WebKit! :)

Cheers,
Thomas.


sgiehl:
I'd suggest do remove the "Firefox" from the version string, as it would be doubled otherwise when name and version are combined

tzi:
Right! 
It's done ;)

PR #5352, opened 2015-06-09
minor correction to detection of NEC devices

matt-kv:
It also prevents user agents containing e.g. "OfficeLiveConnector.1.3" to be wrongly detected as Nec.

### DIFF
--- a/regexes/device/mobiles.yml
+++ b/regexes/device/mobiles.yml
@@ -1929,7 +1929,7 @@ M.T.T.:
 
 # nec
 NEC:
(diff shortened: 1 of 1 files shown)


sgiehl:
That should not be the case. See https://github.com/piwik/device-detector/blob/master/Parser/ParserAbstract.php#L164 which is used to check the regexes. So "Connector" should not match for "NEC".

matt-kv:
Indeed... I had only checked the regex against a list of UAs.

Anyway the major reason is to make the generic NEC regex coherent with the model specific regex. Is it a good idea or should we expect UAs beginning with "NEC" without the "-", thus matching a NEC smartphone device but no particular model ?

sgiehl:
maybe the best would be to do something like  ```NEC[ _\-]```

matt-kv:
no problem.

PR #5360, opened 2015-07-11
Regex for HTC One M8 and OnePlus One

ercpe:


sgiehl:
Would you mind adding tests for the new detections?

ercpe:
Not at all.

PR #5361, opened 2015-07-13
Skip bot detection

phoenix-mstu:
I use device-detector to switch between mobile and desktop version of my website based on visitor's user-agent. Google bot also visits my website using different user agents - mobile and desktop. So I need somehow to switch off the bot detection to show a valid page to Google Bot (otherwise the bot is ever detected as "not mobile" client).


sgiehl:
Thanks for that PR. Guess that might be useful for others, aswell.

PR #5362, opened 2015-07-25
Regex and tests for Sony XPERIA ZR C5503

ercpe:


regexes/device/mobiles.yml
```diff
@@ -1112,6 +1112,8 @@ Sony:
       model: 'Xperia SP'
     - regex: '(?:Sony(?:Ericsson)?)?C5502'
       model: 'Xperia ZR'
+    - regex: '(?:Sony(?:Ericsson)?)?C5503'
```
sgiehl: would you mind combining this with the regex before?

ercpe:
Combined regexes in 6782f3e

sgiehl:
merci

PR #5366, opened 2015-08-16
Add Stringer user-agent regex

swanson:
The full user agent for [Stringer](https://github.com/swanson/stringer) (a self-hosted RSS reader) is:

``` ruby
USER_AGENT = "Stringer (https://github.com/swanson/stringer)"
```

We don't report a version - is it alright to omit it from the yml file here?


sgiehl:
Would you mind adding a test?

swanson:
Tests added, squashed into the other commit.

PR #5386, opened 2015-12-09
Add detection for Coc Coc browser

hongquan:
Homepage: http://coccoc.vn/en/

This is 1 of the two most popular browsers in Viet Nam (the other one is Chrome).


sgiehl:
Thx for the PR :+1:

PR #5423, opened 2016-01-30
Lots of Podcatchers from podseed.org

fstern:
Hi, 

I went through 3 years worth of podseed.org logs and added most Podcast Clients to the "mediaplayers" file and also a few other clients I found.


sgiehl:
Need to have a closer look at this PR to ensure all added regexes are correct and in the correct "category"

Would you mind adding tests for the newly added detections? Or maybe you send a list of useragents to stefan@piwik.org. Then I'll add the tests on my own to validate the regexes.

fstern:
Hi Stefan, 

sure, I'll add some tests for the Useragents. I thought the category "mediaplayer" would be appropriate for most podcatchers.

sgiehl:
I guess as long as the podcatchers are local tools, the category mediaplay should be the best match.
Online services fetching podcasts without any user action should maybe better be added as a bot, as I don't think they should be detected as a real client.

sgiehl:
Do you have time to rebase your pull request and add some tests?
Otherwise I'll start to add your changes step by step.

fstern:
Hi Stefan, 

Sorry, it's delayed, mostly due to time constraints and real life taking over. 

Best, 

Falk

On Thu, Feb 18, 2016 at 01:31:28PM -0800, Stefan Giehl wrote:
> Do you have time to rebase your pull request and add some tests?
> Otherwise I'll start to add your changes step by step.
-- 
The X approach to device independence is to treat everything like a
MicroVAX framebuffer on acid. - The Unix Haters Handbook on X11

sanchezzzhak:
I might be able to update this
@sgiehl do we need these changes?

sgiehl:
@sanchezzzhak guess it at least makes sense to add those we have example useragents for...

PR #5435, opened 2016-02-17
Remove unused param

emirb:


sgiehl:
That "unused" parameter is to be consistent with doctrine/cache (https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/Cache.php#L79)

emirb:
What's the purpose then, if Doctrine's `Cache` interface isn't implemented in `DeviceDetector\Cache\StaticCache`?

ThaDafinser:
@emirb so Doctrine\Cache is a optional dependency and is not required, but can be used.

See here https://github.com/piwik/device-detector/blob/master/DeviceDetector.php#L747-L751

emirb:
@ThaDafinser Thanks for clarifying. I agree it's not a clean solution, just my 5 cents.

sgiehl:
If someone has the time to make a cleaner solution. Feel free to issue a PR for that.

PR #5455, opened 2016-02-29
Fix problem with reusing an instance of device detector for multiple detections

sgiehl:
In some special cases the parse result of device detector wasn't reset correctly when multiple detection were done with one instance


ThaDafinser:
:+1:

PR #5491, opened 2016-04-21
Use configured cache for vendor fragment parsing

sideshowcecil:
The configured cache implementation is injected into each used parser instance but this injection is missing when parsing vendor fragments.


sgiehl:
Thx.

PR #5496, opened 2016-04-27
Fixed README and removed debugging statement from Symfony YAML integration

cundd:
I wanted to use the Symfony YAML integration but found a debugging statement there. Is the integration unstable, or was the output "test" simply a mistake?


sgiehl:
Huh, that was a mistake

PR #5508, opened 2016-06-10
Add Wechat browser

fengerzh:
As Wechat used more frequently in China, we should include this browser in, otherwise it will show as Unknown Browser.

### DIFF
--- a/regexes/client/browsers.yml
+++ b/regexes/client/browsers.yml
@@ -24,6 +24,13 @@
   engine:
     default: 'WebKit'
 
+# Wechat Phone Browser
+- regex: 'MicroMessenger(?:/(\d+[\.\d]+))?'
+  name: 'Wechat Browser'
+  version: '$1'
+  engine:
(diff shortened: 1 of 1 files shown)


sgiehl:
Any reason why the detection as mobile app is wrong?
See https://github.com/piwik/device-detector/blob/master/regexes/client/mobile_apps.yml#L33

fengerzh:
Sorry, did not notice that. But I couldn't find any section in Piwik list out the mobile apps. There are only "Operation System", "Browser Name", "Browser Engine" in the Software page. And in "Browser Name" section, there are lots of "Unknown".

sgiehl:
Ah ok. I see your point. Piwik currently does not everything device detector is able to detect. Piwik will only use detected browsers and will report all other as unknown. We are planning to bring to stuff in Piwik aswell, but that might take some more time...

fengerzh:
Got it. Thank you!

PR #5510, opened 2016-07-14
Add detection for 10+ podcast clients

benzimmer:
This adds support for 10+ podcast clients on different devices/OSes.

Please let me know if there are things that need to be changed for this to get merged.


sgiehl:
I'll have a closer look at that one the coming days...

sgiehl:
I'll have a closer look at that one the coming days...

PR #5519, opened 2016-08-08
fix Intex regular group

sanchezzzhak:
Current regular correct for php but it wrong nodejs/javascript

### DIFF
--- a/regexes/device/mobiles.yml
+++ b/regexes/device/mobiles.yml
@@ -1975,12 +1975,12 @@ INQ:
 
 # Intex
 Intex:
-  regex: 'Intex|(Aqua|Cloud)[ _\.]([^/;\)]+)(?:[ _]Build|(?<!Build)/|\))'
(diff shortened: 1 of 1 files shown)


sgiehl:
nodejs/javascript doesn't support negative lookbehindes? If so the regex needs to be changed completely. Adding a `:` changes the result.

sanchezzzhak:
yes, lookbehindes old different syntax
js `(^(?!filename)$)`
php `(?<!filename)`

test js/php
https://regex101.com/r/xT3xZ2/1

sgiehl:
It also needs to match `Aqua Y2/V1 Linux/3.4.5 Android/4.2.2 Release/03.26.2013 Browser/AppleWebKit534.30 Mobile Safari/534.30 MBBMS/2.2;`

sgiehl:
I'll close this one now. Feel free to reopen a pull request if you solved the problem.

PR #5527, opened 2016-08-13
Epiphany has rebraneded as GNOME Web

### LINKED ISSUE #5524: Epiphany has rebraneded to "GNOME Web"
The browser formerly known as Epiphany has rebraneded to “GNOME Web”. The User-Agent remains the same, but the display name should be updated.

da2x:
Resolves #5524.

### DIFF
--- a/regexes/client/browsers.yml
+++ b/regexes/client/browsers.yml
@@ -387,7 +387,7 @@
 
 #Epiphany
 - regex: 'Epiphany(?:/(\d+[\.\d]+))?'
-  name: 'Epiphany'
+  name: 'GNOME Web'
   version: '$1'
   engine:
     default: 'Gecko'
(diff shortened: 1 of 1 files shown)


regexes/client/browsers.yml
```diff
@@ -387,7 +387,7 @@
 
 #Epiphany
 - regex: 'Epiphany(?:/(\d+[\.\d]+))?'
-  name: 'Epiphany'
+  name: 'GNOME Web'
```
sgiehl: Please also update the comment above, and the name in the list of available browsers

sgiehl:
You may also need to rename it here https://github.com/piwik/device-detector/blob/master/Parser/Client/Browser.php#L70 in order to get the tests running again.

PR #5528, opened 2016-08-13
Added BriskBard

### LINKED ISSUE #5525: Add BriskBard browser
[BriskBard](https://www.briskbard.com/) is a new browser built on Trident, but it also has it’s own proprietary layout engine (also called BriskBard).

User-Agent: `BriskBard/1.0 (Windows 10) BriskBard/1.0`

da2x:
Added [BriskBard browser](https://www.briskbard.com/).
Resolves #5525.

### DIFF
--- a/regexes/client/browsers.yml
+++ b/regexes/client/browsers.yml
@@ -919,6 +919,11 @@
   name: 'BlackBerry Browser'
   version: ''
 
+#BriskBard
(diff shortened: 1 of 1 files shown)


sgiehl:
Thx for the PR. Would you please add the new browser in the list of available browsers in https://github.com/piwik/device-detector/blob/master/Parser/Client/Browser.php#L29
and please also add a test to proof the detection works

PR #5532, opened 2016-08-25
Isivioo and B-Line browsers detection

wazelin:
http://www.blineapp.com/
http://www.isivioo.com/


sgiehl:
would you mind resolving the conflicts? Afterwards I'll merge this on.

PR #5536, opened 2016-08-31
Fix warnings with regexps that present in Ruby

mkdynamic:
This PR fixes some warnings about the regular expressions that present when using some patterns with Ruby via https://github.com/podigee/device_detector.

I believe this does not change the semantics of those patterns -- so even though the motivation for this PR is somewhat external, I am hoping it will be merge-able for the benefit of the `device_detetor` Rubygem.

Let me know if you have any questions or I am make further changes to help make this PR merg-eable.

Here is the related change in the PR for that Rubygem repo:
https://github.com/podigee/device_detector/pull/22/files#diff-e4f032e0894d753fbcba03794958c7f1L1444

Thanks!

### DIFF
--- a/regexes/device/mobiles.yml
+++ b/regexes/device/mobiles.yml
@@ -1459,7 +1459,7 @@ Sony:
     - regex: '(?:Sony-)?(KDL?-?[0-9a-z]+)'
       model: '$1'
       device: 'tv'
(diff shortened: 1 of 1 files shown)


mkdynamic:
Thanks @sgiehl.

PR #5553, opened 2016-09-28
Improve device detection

DaimonPl:
https://github.com/piwik/device-detector/issues/5543
https://github.com/piwik/device-detector/issues/5548


regexes/device/mobiles.yml
```diff
@@ -2716,6 +2741,17 @@ Mediacom:
...
+MediaTek:
+  regex: 'MT[68]\d{3}[_);/ ]'
+  models:
+    - regex: '(MT8\d{3})'
+      model: '$1'
+      device: 'tablet'
+    - regex: '(MT\d{4})'
+      model: '$1'
+      device: 'smartphone'
+
```
sgiehl: Already thought about adding MediaTek a few times before. But I'm still unsure if those are real devices. For me it seems MediaTek is a vendor for mobile chipsets. See http://www.mediatek.com/en/products/mobile-communications/smartphone1/

regexes/device/mobiles.yml
```diff
@@ -1732,7 +1751,7 @@ Vizio:
 
 # GOCLEVER
 GOCLEVER:
-  regex: 'GOCLEVER|QUANTUM|ARIES|INSIGNIA'
+  regex: 'GOCLEVER|QUANTUM|ARIES|INSIGNIA|ORION_|ELIPSO'
```
sgiehl: Maybe better use `ORION` instead of `ORION_` as not all of those devices needs to have that `_` in the useragent.
For example: `Mozilla/5.0 (Linux; Android 4.2.2; ORION7oL Build/JDQ39) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/35.0.1916.141 Safari/537.36`
Also as far as I had seen all `Orion` devices are tablets, so maybe you could add that separately instead of in the detection below:
```
- regex: 'ORION[ _-]?([^;/]+) Build'
  model: 'Orion $1'
  device: 'tablet'
```
You could use the useragent above to prove that.

regexes/device/mobiles.yml
```diff
@@ -892,9 +904,12 @@ Casio:
 
 # Cat
 Cat:
-  regex: 'Cat ?(tablet|stargate|nova)'
+  regex: 'Cat ?(tablet|stargate|nova)|B15Q'
   device: 'tablet'
   models:
+    - regex: '(B15Q)'
+      model: '$1'
```
sgiehl: Wouldn't use any expressions here, simply:
```
- regex: 'B15Q'
  model: 'B15Q'
```
Won't change anything, but is easier to read (and slightly faster)

regexes/device/mobiles.yml
```diff
@@ -385,8 +388,8 @@ Acer:
       model: 'Liquid Z$1'
     - regex: 'Android.*Z160 Build'
       model: 'Liquid Z4'
-    - regex: 'Android.*Z500 Build'
-      model: 'Liquid Z500'
+    - regex: 'Android.*(Z5\d{2}) Build'
+      model: 'Liquid $1'
```
sgiehl: I'd suggest to have it like this:
```
- regex: 'Android.*Z(5\d{2}) Build'
  model: 'Liquid Z$1'
```
That would prove the `Z` is always a capital letter, even if in the useragent it is lower case

regexes/device/mobiles.yml
```diff
@@ -2155,17 +2180,17 @@ Kazam:
...
 # Kiano
 Kiano:
-  regex: 'Kiano|Elegance_5_0|CORE 10.1 DUAL 3G|Intelect103G|SlimTab7[_ ]3GR[);/ ]'
+  regex: 'Kiano|Elegance_5_0|CORE 10.1 DUAL 3G|Intelect ?(7|8|10)|Slim ?Tab ?\d+(?:[_ ]3GR?)?[);/ ]'
   device: 'smartphone'
   models:
-    - regex: 'SlimTab7[_ ]3GR'
-      model: 'SlimTab7 GR'
+    - regex: 'Slim ?Tab ?(\d+)(?:[_ ]3GR?)?'
+      model: 'Slim Tab $1'
```
sgiehl: Do you ignore the `3GR` on purpose? Or should we use `Slim ?Tab ?(\d+)([_ ]3GR?)?` / `Slim Tab $1 $2` so it would be added if available?

sgiehl (changes requested):
Finally found a bit of time to make a short review. Please have a look at the comments.

regexes/device/mobiles.yml
DaimonPl: Updated

regexes/device/mobiles.yml
DaimonPl: updated

regexes/device/mobiles.yml
DaimonPl: Matching without `_` wrongly detects following user agent from other tests:

`Linux; U; Android 2.3.6; it-it; NGM Orion Build/GRK39F) AppleWebKit/533.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/533.1`

I'd rather leave it as it's now as I don't have idea how to target this case

regexes/device/mobiles.yml
DaimonPl: Thought that it's still better detection than no detection. let me know :)

regexes/device/mobiles.yml
```diff
@@ -4789,6 +4833,9 @@ Toshiba:
     - regex: 'AT(100|200|270|300|330|374|400|470|500|503|570|703|830)'
       model: 'Regza AT$1'
       device: 'tablet'
+    - regex: 'AT(10-A)'
+      model: 'Excite AT$1'
```
DaimonPl: done

DaimonPl:
Any update for this one? :)

regexes/device/mobiles.yml
sgiehl: Don't think it's important. Maybe we could/should remove that in some other places, as well

sgiehl:
Added another comment. As soon as you have changed that. I'll merge the PR.
Cheers and sorry for the long delay...

regexes/device/mobiles.yml
DaimonPl: removed

DaimonPl:
Removed mediatek

sgiehl (approved):

PR #5556, opened 2016-10-28
add dorodo browser device nokia

sanchezzzhak:
test regex https://regex101.com/r/njV2Nn/1


sgiehl:
The new browser name will also be needed to be defined here: https://github.com/piwik/device-detector/blob/master/Parser/Client/Browser.php#L68
Please also add some tests.

sgiehl:
I'll merge this now, and add a test afterwards. Thx for the Pull Request :+1:

PR #5559, opened 2016-11-15
Fixed typo for Backlink-Check.de robot

King2500:
*see changes*


sgiehl:
Seems there is a test for the "broken" detection. would you mind fixing that, as well?

sgiehl:
merci :+1:

PR #5584, opened 2017-02-10
Added Traackr.com

shaeney:
HI, I am not really familiar with Git or php, being a .Net developer who uses TFS.

But I'd like to contribute


Findus23:
@shaeney Could you add tests in `Tests/fixtures/bots.yml` for the new bots? (Similar to https://github.com/piwik/device-detector/commit/c219415c2ad114874adb77ad6753d1663710c6b7)

shaeney:
Can anyone help? I am not familiar with open source/git, so dont know how to find out what I need to do to fix this?

Findus23:
@shaeney The error travis is reporting (https://travis-ci.org/piwik/device-detector/jobs/201858723#L367) is that in https://github.com/piwik/device-detector/pull/5584/files#diff-658652a1b8930dae0035e1dee2883900R1390 you set the url of the bot to an empty string but in `Tests/fixtures/bots.yml` you are missing the `url` key. 
Best solution is to add an url (http://feeltiptop.com) to both of them.

In addition you'll need a similar test for Traackr.com and WebClient.
And it would be nice to add more information to WebClient if you have any. (Or just add it with the name "Generic Bot" if there are none)

shaeney:
@Findus23 Can you give me a hand? While I am familiar with TDD in a .Net world, I dont really understand whats expected of me here. I've tried to reverse engineer what I need to do, but cant work it out?

I thought Id need to add text to
Tests\fixtures\bots.yml for example

- 
  user_agent: Traackr.com
  bot:
    name: Traackr.com
    category: Crawler
    url: http://www.traackr.com

and this would be matched by the equivalent in the regex bots.yml?

But I cant interpret what the error log is telling me?

Findus23:
@shaeney The last commit is looking good. Are you sure that the complete user agent is `Traackr.com`? 
For the key `user_agent` in the test you need an example of the user agent.

Tests/fixtures/bots.yml
```diff
@@ -2649,4 +2649,35 @@
...
+  user_agent: http://feeltiptop.com
+  bot:
+    name: feeltiptop.com
+    category: Crawler
+    producer:
+      name: feeltiptop.com
+      url: http://feeltiptop.com
+
+- 
+  user_agent: Traackr.com
```
Findus23: This test is not needed any more as the second one is working

regexes/bots.yml
```diff
@@ -1376,10 +1376,27 @@
...
 
+- regex: 'Traackr.com'
+  name: 'Traackr.com'
+  category: 'Crawler'
+  url: 'http://www.traackr.com'
+
+- regex: 'feeltiptop'
+  name: 'feeltiptop.com'
+  category: 'Crawler'
+  url: ''
```
Findus23: Could you add the url here instead of an empty string?

Tests/fixtures/bots.yml
```diff
@@ -2649,4 +2649,35 @@
...
   bot:
-    name: Generic Bot
\ No newline at end of file
+    name: Generic Bot
+
+- 
+  user_agent: http://feeltiptop.com
+  bot:
+    name: feeltiptop.com
+    category: Crawler
```
Findus23: You need to add the url here

regexes/bots.yml
```diff
@@ -1376,10 +1376,27 @@
...
+  producer:
+    name: 'feeltiptop.com'
+    url: 'http://feeltiptop.com'
+
+- regex: 'WebClient/1.0'
+  name: 'WebClient'
```
Findus23: You need to change the name to 'Generic Bot' to match the test.

Findus23 (changes requested):
I hope this helps

regexes/bots.yml
```diff
@@ -1408,6 +1408,27 @@
...
+
+# Some libraries used for bots
+- regex: '(HTTP_Request2|Mechanize|aiohttp|Google-HTTP-Java-Client|WWW-Mechanize|faraday)'
+  name: 'Generic Bot'
```
sgiehl: Guess this block moved in while merging. It shouldn't be present in that file anymore, so please remove it as well.

sgiehl:
Closing stalled pull request

PR #5587, opened 2017-02-15
A few more Bots: DuckDuckGo Bot, "checker", Guzzle, PocketParser 

Findus23:
Sorry for the multiple pull requests, but I keep finding new bots.
 - DuckDuckGo Bot
 - "checker" as generic bot suffix
 - Guzzle (PHP HTTP client)
 - PocketParser (I added the new category 'Read-it-later Service')
I also wanted to add Instapaper, but they seem to use a not helpful user-agent: 
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_3) AppleWebKit/534.55.3 (KHTML, like Gecko) Version/5.1.3 Safari/534.53.10`

### DIFF
--- a/Tests/Parser/Client/fixtures/library.yml
+++ b/Tests/Parser/Client/fixtures/library.yml
@@ -82,3 +82,39 @@
     type: library
     name: Guzzle (PHP HTTP Client)
     version: "3.9.3"
+-
+  user_agent: HTTP_Request2/2.3.0 (http://pear.php.net/package/http_request2) PHP/5.3.3
+  client:
+    type: library
+    name: HTTP_Request2
+    version: 2.3.0
+-
+  user_agent: Mechanize/2.7.3 Ruby/1.9.3p551 (http://github.com/sparklemotion/mechanize/)
+  client:
+    type: library
+    name: Mechanize
+    version: 2.7.3
+-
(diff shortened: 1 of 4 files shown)


sgiehl:
Wouldn't be `Read-it-later Service` similar to the category `Feed Fetcher`

Findus23:
I'm not sure. 
In my opinion they don't automatically fetch feeds, but rather crawl single articles on user action.

Findus23:
Sorry, I didn't now that Guzzle was already added and then moved to regexes/client/libraries.yml
https://github.com/piwik/device-detector/blob/master/regexes/client/libraries.yml#L12

Should I also move the other libraries there?
https://github.com/piwik/device-detector/pull/5587/files#diff-658652a1b8930dae0035e1dee2883900R1395

Is it useful to assume that every request using the default user agent of a http library is a bot?
If yes, should the bot detection also include the regexes from regexes/client/libraries.yml?

sgiehl:
I wouldn't assume that every HTTP library is a bot. There might be applications or apps using such HTTP libraries and shouldn't be detected as bots.

Findus23:
I have now removed the libraries from the bot detection and added them to the libraries detection (with new tests)

PR #5593, opened 2017-02-27
fixed detect Asus zen 2 & go; + add zen 3 model + new brands

sanchezzzhak:
fix detect Asus zen 2 & go; and add zen 3 model
new brand Supra
new brand MIXC
new brand Homtom
add model Samsung SM1S


Parser/Device/DeviceParserAbstract.php
```diff
@@ -223,6 +224,7 @@
         'ME' => 'Metz',
         'MX' => 'MEU',
         'MI' => 'MicroMax',
+        'MIX' => 'MIXC',
```
sgiehl: Please use only 2 characters as shortcode. If useful shortcodes are already taken use something like 'M1', 'M2'
(Piwik isn't able to handle longer short codes)

regexes/device/mobiles.yml
```diff
@@ -715,10 +715,16 @@ Asus:
...
-    - regex: '(?:ASUS_)?Z00[SV]D'
+    - regex: '(?:ASUS_)?(Z00[SV]D|X009[DB][DA])'
       model: 'ZenFone Go'
+    - regex: '(?:ASUS_)?X013D'
+      model: 'ZenFone Go Life'
+    - regex: '(?:ASUS_)?(X008)'
+      model: 'ZenFone 3 Pegasus'
     - regex: '(?:ASUS_)?Z00UD'
       model: 'ZenFone Selfie'
+   - regex: '(?:ASUS_)?(Z017DA|ZE520KL)
```
sgiehl: this line lacks a `'` in the end and breaks some detections

sgiehl:
Thanks for the Pull Request. I'll merge this one in a new branch and fix the tests before merging into master.

PR #5594, opened 2017-03-02
Add new brand SENSEIT and detect many models 

sanchezzzhak:
new brand SENSEIT detect models R390, R450, L301, A200, A109, E510, E500, E400


sanchezzzhak:
When it will be possible, can you please merge the request?

regexes/device/mobiles.yml
```diff
@@ -1540,6 +1539,9 @@ Sony:
       model: 'Xperia X'
     - regex: '(?:Sony(?:Ericsson)?)?F5321'
       model: 'Xperia X Compact'
+    - regex: 'E5506'
```
sgiehl: Could you please change that to the previous regex: `'(?:Sony(?:Ericsson)?)?E5506'`

regexes/device/mobiles.yml
```diff
@@ -1540,6 +1539,9 @@ Sony:
...
+      model: 'Xperia C5 Ultra'
+      brand: 'Sony'
```
sgiehl: no need to define the brand here, as `Sony` is the default. You can remove this line

regexes/device/mobiles.yml
```diff
@@ -1267,9 +1267,17 @@ Rikomagic:
   device: 'tv'
   model: 'MK$1'
 
+# Senseit
+Senseit:
+   regex: 'SENSEIT[ _]?([^;/)]+) Build'
+   device: 'smartphone'
+   models:
+     - regex: '(R390|R450|E[54][10]0|A109|A200|L301)'
```
sgiehl: Why are you defining a sub regex here? The global regex should already match everything in this case. You could directly use:

```
# Senseit
Senseit:
   regex: 'SENSEIT[ _]?([^;/)]+) Build'
   device: 'smartphone'
   model: '$1'
```

It can be changed as soon as some extra detections needs to be added

sgiehl (changes requested):

regexes/device/mobiles.yml
sanchezzzhak: I did not know that it was possible, i will correct regex

regexes/device/mobiles.yml
sanchezzzhak: done

regexes/device/mobiles.yml
sanchezzzhak: done

sgiehl:
Seems correct now. Will merge as soon as the tests passed...

sgiehl (approved):

PR #5601, opened 2017-03-28
Adjust Google-HTTP-Java-Client regex to be compatible with ruby

seandilda:
All tests still pass, this should just make the regexes more portable

### DIFF
--- a/regexes/client/libraries.yml
+++ b/regexes/client/libraries.yml
@@ -55,7 +55,7 @@
   name: 'aiohttp'
   version: '$1'
 
-- regex: 'Google-HTTP-Java-Client(?:/(\d+[\.\d-\w]+))?'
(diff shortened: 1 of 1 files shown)


Findus23:
As I wrote that regex and am quite a beginner, I would like to ask:
What is the difference? Why is the other version not compatible with ruby?

seandilda:
@Findus23, since it is in a [],  ruby sees the '\d-\w' as attempting a range from \d to \w (akin to doing [a-z]).  It doesn't know how to process that.  However, if you make the last character in the [] a -, it knows you aren't trying to create a range, but instead just want - to be one of the characters it matches.

Findus23:
@seandilda 
Ah, that makes sense. I didn't think of that.

PR #5625, opened 2017-06-05
Split Firefox into Firefox and Firefox Mobile

### LINKED ISSUE #5621: Split Firefox into mobile and desktop (similar to Chrome)
from https://forum.piwik.org/t/piwik-firefox-user-agent/24463/

All Firefox Browsers are just called "Firefox" (https://github.com/piwik/device-detector/blob/master/regexes/client/browsers.yml#L177) and as a result are grouped as one browser in Piwik.

On the other hand Chrome is split up in "Chrome" and "Chrome Mobile" (https://github.com/piwik/device-detector/blob/master/regexes/client/browsers.yml#L462).

We should decide which one is correct and change the other one.

Findus23:
fixes #5621

The change is easy, but I am not sure how to handle the update, as this changes Piwik statistics quite a bit.


sgiehl:
There is no possibility to update anything in Piwik, as the useragents aren't stored.
So it might screw up some statistics like row evolution for Firefox a bit, but we can't do anything about it.

PR #5627, opened 2017-06-06
Added detection for new Samsung Galaxy tablets

derastronaut:
I added some Samsung-devices which were just shown like SM-G0123 or so in Piwik.


sgiehl:
Could you please update the tests, as well? I'll merge as soon as the tests passes...

PR #5651, opened 2017-09-30
Adds detection for TrekStor devices

sgiehl:
@Findus23  same here ;)


Findus23:
again added in https://github.com/piwik/piwik-icons/commit/8b50f17fb4d7cfc43bc010d04b1df7539187e905

Fascinating how many brands I have never heared of

sgiehl:
> Fascinating how many brands I have never heared of

I'm going through a long list of user agent at the moment and checking if they are detected correctly... guess there will be some brands

Issue #5646, opened 2017-09-01
Facebook messenger detection


etienne-martin:
Facebook messenger is detected as the facebook mobile app. However, the version number displayed is that of facebook messenger.

```javascript
"client": {
    "type": "mobile app",
    "name": "Facebook",
    "version": 132 // <-- Facebook messenger version number
}
```

Here are two user agents from messenger, one on iOS and the other on Android. Both contain the string `messenger`:

```
Mozilla/5.0 (iPhone; CPU iPhone OS 10_3_3 like Mac OS X) AppleWebKit/603.3.8 (KHTML, like Gecko) Mobile/14G60 [FBAN/MessengerForiOS;FBAV/132.0.0.41.90;FBBV/69171754;FBDV/iPhone8,4;FBMD/iPhone;FBSN/iOS;FBSV/10.3.3;FBSS/2;FBCR/Koodo;FBID/phone;FBLC/en_US;FBOP/5;FBRV/0]
Mozilla/5.0 (Linux; Android 7.1.2; Nexus 5X Build/N2G47F; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/57.0.2987.132 Mobile Safari/537.36 [FB_IAB/MESSENGER;FBAV/114.0.0.21.71;]
```

Here's a long list of user agents gathered from facebook messenger: https://gist.github.com/etienne-martin/bea6b38c6999c720d91bd77010fae4f4

I would like to add support for these user agents but I suck at regex.

Can someone help me with that?

sanchezzzhak:
```
FBAN: FaceBook Application Name
FBAV: FaceBook Application Version
FBBV: FaceBook Build Version
FBDV: FaceBook Device Version
FBMD: FaceBook Major Device
FBSN: FaceBook System Name
FBSV: FaceBook System Version
FBSS: FaceBook System Something :)
FBCR: FaceBook CarrieR
FBID: FaceBook Identity of Device
FBLC: FaceBook Language Code
```

etienne-martin:
Here's the regex I came up with:

```yml
# Facebook Messenger
- regex: '(?:MessengerForiOS|MESSENGER).(?:FBAV)(?:[ /]([\d\.]+))?'
  name: 'Facebook Messenger'
  version: '$1'
```

I tested it with these two user agents and it seems to work fine.

```
[...]

PR #5655, opened 2017-09-30
Facebook messenger detection #5646

etienne-martin:
Here's my attempt at detecting Facebook Messenger.


sgiehl:
thanks for the PR :+1:

Issue #5609, opened 2017-04-19
Additional browsers


Findus23:
Now that we are using https://github.com/alrra/browser-logos/tree/master/src for the browser icons, we may add support for some of the browsers they have icons for.

Unfortunatly many of them don't have an unique user agent, but I'll see what I can find.

sgiehl:
Could you maybe add the full list here, so we can check what we have - and what we need to investigate?

Findus23:
# Current Browsers
## [adblock browser](https://adblockbrowser.org/)
- [ ] added
- seems to have no unique user agent

## [Beaker](https://beakerbrowser.com/)
- [ ] added

## [Blisk](https://blisk.io/)
- [ ] added

## [Boat](http://www.boatmob.com/mobile/product_browser_free.html)
- [ ] added

## [Browser.html](https://github.com/browserhtml/browserhtml#readme)
- [ ] added

## [Cent](https://www.centbrowser.com/)
- [ ] added

## Chrome Beta/Dev/Canary
- probably don't need to be detected as seperate browsers
- maybe check if they are detected as Chrome

## [Coast](https://www.opera.com/mobile/coast)
- [ ] added

## [Cốc Cốc](https://coccoc.com/)
- [ ] added
- "among the five most popular browsers in Vietnam"
- user agents similar to 
    - `Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) coc_coc_browser/45.0 Chrome/39.0.2171.98 Safari/537.36`
    - `Mozilla/5.0 (Windows NT 5.1) AppleWebKit/537.36 (KHTML, like Gecko) coc_coc_browser/42.0 CoRom/36.0.1985.146 Chrome/36.0.1985.146 Safari/537.36`

## [Dooble](http://dooble.sourceforge.net/)
- [ ] added

## [Epic](https://www.epicbrowser.com/)
- [ ] added

## Firefox Beta/Developer Edition/Aurora/Nightly
- as Chrome

## [Firefox Focus](https://github.com/mozilla-mobile/focus)
- [ ] added

## [Iridium](https://github.com/iridium-browser)
- [ ] added

## [MIHTool](http://www.mihtool.com/)
- [ ] added
- not really a browser

## [Min](https://palmeral.github.io/min/)
- [ ] added


## [NetSurf](http://www.netsurf-browser.org/)
- [ ] added

## [NW.js](https://nwjs.io/)
- [ ] added
[...]

PR #5660, opened 2017-10-01
Adds detection for Firefox Focus browser

sgiehl:
ping @Findus23


Findus23:
So it can be ticked from #5609

PR #5665, opened 2017-10-03
PSR-6 & PSR-16 compliant cache bridges

bpolaszek:
Hello there,

PHP-Fig recommends 2 standards of caching:

* [PSR-6](http://www.php-fig.org/psr/psr-6/) which covers the usage of a Cache Pool
* [PSR-16](http://www.php-fig.org/psr/psr-16/) which provides a more simple caching contract.

There are already lots of implementations of these new standards, but none of them is currently compatible with **DeviceDetector** because it relies on a home-made Cache and allows a commonly used, but non standard, _Doctrine Cache_.

This PR creates a bridge for these types of caching, and allow any implementation of PSR-6 and PSR-16 with this library, and features like:
* Cache prefixing to avoid key collisions
* Dozens of adapters such as:
    * Array
    * Apcu
    * Memcached
    * Redis
    * Doctrine
    * Filesystem
    * Pdo
    * NullCache (null pattern)
    * Chain (multiple cache handlers)
    * ...

Example use with the Symfony PSR-6 implementation:
```php
require_once __DIR__ . '/vendor/autoload.php';

use DeviceDetector\Cache\PSR6Bridge;
use Symfony\Component\Cache\Adapter\RedisAdapter;

$redis = new Redis();
$redis->connect('127.0.0.1');
$cache = new RedisAdapter($redis, 'device_detector');
$deviceDetector = new DeviceDetector();
$deviceDetector->setCache(
    new PSR6Bridge($cache)
);
```

Example use with the Symfony PSR-16 implementation:
```php
require_once __DIR__ . '/vendor/autoload.php';

use DeviceDetector\Cache\PSR16Bridge;
use Symfony\Component\Cache\Simple\FilesystemCache;

$cache = new FilesystemCache('device_detector', $ttl = 20);
$deviceDetector = new DeviceDetector();
$deviceDetector->setCache(
    new PSR16Bridge($cache)
);

```

Cf. https://github.com/piwik/device-detector/issues/5572#issuecomment-273423924


bpolaszek:
Travis fails on PHP 5.4 - that's because I `require --dev symfony/cache` for testing the interfaces, which requires PHP >= 5.5. Don't know what to do :confused:

bpolaszek:
> Travis fails on PHP 5.4 - that's because I `require --dev symfony/cache` for testing the interfaces, which requires PHP >= 5.5. Don't know what to do :confused:

Commit amended: I now `require --dev matthiasmullie/scrapbook`, which is another PSR-6 / PSR-16 implementation, compatible from PHP 5.3+. 

All tests pass! Which proves you can safely rely on any PSR-6 / PSR-16 implementation! :tada:

sgiehl:
Thanks for the PR!
Would you mind updating the [README](https://github.com/piwik/device-detector/blob/master/README.md#caching) and adding examples there as well?

bpolaszek:
Hello @sgiehl,

Of course, I'll do this ASAP!
Just one question, because all these cache things are becoming a little bit messy (and thus, messy to explain):

* We can implement the library's built-in cache interface
* We can use a Doctrine cache bridge
* We can implement (or use a library that implements) PSR-6
* We can implement (or use a library that implements) PSR-16

Do you plan on dropping Doctrine Cache and built-in cache interface in the next major version (v4)? 

Then we'll have 2 choices:
* Accept both PSR-6 and PSR-16 caches (with a built-in ArrayCache fallback, maybe)
* Define which one the library relies on. This allows `CacheInterface`or `CacheItemPoolInterface` type-hinting in classes that need cache. Then I would suggest PSR-16, since we need something quite simple.

This way the library would rely on PSR standards, and this would shorten the docs on that point.

What's your opinion?

sgiehl:
We could drop Doctrine support in next major release. Removing the built-in cache interface would only be an option if we adjust Piwik first. It uses https://github.com/piwik/component-cache for caching and implements the cache interface. See https://github.com/piwik/piwik/blob/3.x-dev/core/DeviceDetectorCache.php

bpolaszek:
OK, for now I won't mention anything of 4.x in the README. Since Caching is a shared component, it involves many things I don't control at the moment (I only use DeviceDetector and not the whole Piwik project).

I have committed a new README after rebasing on master. Tell me what do you think about it? I removed the line about the "cache redesign", since it's been done for almost 3 years.

sgiehl:
@bpolaszek  would you mind rebasing on master another time? Changes to README are fine, would merge afterwards.

bpolaszek:
Hello @sgiehl,

Done.

Thank you!
Ben

PR #5667, opened 2017-10-05
Added support for Edge/Android and Edge/iOS ...

nickurt:
I used the following shortcodes to referring to Edge on a mobile-device
```
'PB' => 'Microsoft Edge for Android'  // Project Spartan on Blink-engine
'PI' => 'Microsoft Edge for iOS' // Project Spartan on iOS
```
https://blogs.windows.com/msedgedev/2017/10/05/microsoft-edge-ios-android-developer/

### DIFF
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -151,6 +151,8 @@ class Browser extends ClientParserAbstract
         'PO' => 'Polaris',
         'PT' => 'Polarity',
         'PS' => 'Microsoft Edge',
+        'PB' => 'Microsoft Edge for Android',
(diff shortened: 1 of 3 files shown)


sgiehl:
I'm not sure if we should add it that way. Maybe we should avoid to have the operating system name in any browser name. Might be better to detect all as 'Microsoft Edge' instead.

(Side note: Maybe we should also combine `Chrome Mobile` and `Chrome Mobile iOS`)

nickurt:
Updated the pull request with using the same Edge-naming now ...

sgiehl:
Thanks for the update

PR #5670, opened 2017-10-11
Update bots.yml

razseo:
Improve detect seznan images mail proxy.

### DIFF
--- a/regexes/bots.yml
+++ b/regexes/bots.yml
@@ -500,7 +500,15 @@
   producer:
     name: 'Google Inc.'
     url: 'http://www.google.com'
-
+    
+- regex: 'Seznam Email Proxy'
+  name: 'SeznamEmailProxy'
+  category: 'Crawler'
+  url: ''
+  producer:
+    name: 'Seznam.cz, a.s.'
+    url: 'http://www.seznam.cz/'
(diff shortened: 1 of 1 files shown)


regexes/bots.yml
```diff
@@ -500,7 +500,15 @@
   producer:
     name: 'Google Inc.'
     url: 'http://www.google.com'
-
+    
+- regex: 'Seznam Email Proxy'
+  name: 'SeznamEmailProxy'
```
sgiehl: Guess you switched the content of `regex` and `name`

sgiehl (changes requested):
Could you maybe also add a test for this bot, to prove if works?

regexes/bots.yml
razseo: Mozilla/5.0 SeznamEmailProxy/2.0.174
--


Here is full user agent from my access log.

razseo: About test - you mean unit test? I never write it :(

razseo:
So nobody need this?

regexes/bots.yml
sgiehl: ok. will add it after merging

PR #5673, opened 2017-10-17
Add bots for Czech Republic

soukicz:
One major search engine and three "google shopping" alternatives.

### DIFF
--- a/Tests/fixtures/bots.yml
+++ b/Tests/fixtures/bots.yml
@@ -1908,6 +1908,33 @@
       name: Seznam.cz, a.s.
       url: http://www.seznam.cz/
 -
+  user_agent: Heurekabot-Feed/1.0 (+https://sluzby.heureka.cz/napoveda/heurekabot/)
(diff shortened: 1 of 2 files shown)


sgiehl:
Thanks for the PR. Would you mind adding some tests for those bots in [bots.yml](https://github.com/piwik/device-detector/blob/master/Tests/fixtures/bots.yml) or providing the user agents so we can add tests?

Findus23:
Hi, It would be also nice if you could provide sqare icons for those websites, so I could add them to the piwik-icons.

sgiehl:
@Findus23  are you including images for bots as well?

Findus23:
Ah, thanks.
Forget what I wrote, I thought this was the searchengine-and-social-list
(@soukicz, you may want to add those websites [there](https://github.com/piwik/searchengine-and-social-list) if they are missing)

soukicz:
Tests added

PR #5676, opened 2017-11-02
Some more bots

Findus23:
A few bots from my web server.

I didn't run the tests as they don't seem to be working for me. (But it seems like they are working)


sgiehl:
:+1:

PR #5699, opened 2018-01-15
Bugfix/bot parser discard details

wazelin:


sgiehl:
Thanks for the PR. Guess you're right. When discarding additional bot information, there is no need to walk through all single regex...

PR #5709, opened 2018-02-08
Lenovo tab4 and sentry

weberdever:


regexes/device/mobiles.yml
```diff
@@ -3067,6 +3067,12 @@ Lenovo:
     - regex: 'Yoga Tablet([^;/]+) Build'
       model: 'Yoga Tablet$1'
       device: 'tablet'
+    - regex: 'Lenovo TB\-X704F Build/NMF26F'
+      model: 'Lenovo Tab 4 10 Plus'
```
sgiehl: would you remove the brand name from the model name? The model should only be `Tab 4 10 Plus`.

regexes/device/mobiles.yml
```diff
@@ -3067,6 +3067,12 @@ Lenovo:
...
+      device: 'tablet'
+    - regex: '(Lenovo TB[0-9]?\-[a-z0-9\-]+) Build'
```
sgiehl: Please exclude the brand name from the match by using `(TB[0-9]?\-[a-z0-9\-]+) Build`. There's no need to match Lenovo at all, as this is already done with the base regex for the brand.

sgiehl (changes requested):
Thanks for the PR. Left some comments on stuff that needs to be adjusted.

regexes/device/mobiles.yml
weberdever: +

regexes/device/mobiles.yml
weberdever: \+

sgiehl (approved):

PR #5712, opened 2018-02-19
Adds bot-parsers as injectionable objects

NanneHuiges:
Handles the botparser equivalent to the client/device parsers. This way
users of the library can add their own botparser that might not be
usefull for the library, but are needed in their own situation. (for instance: 
add a botparser that can recognize internal bots with specific user-agents, 
that would never be encountered in the wild and therefore not very usefull 
in the library)

Notes:
* The methods have been made to be as equivalent to their client/device counterparts.
* The `parseBot()` function uses the same logic as the `parsClient()` and `parseDevice()` functions. 
* The `addBotParser()` function is simpeler as there is only a single botparser in the library: we don't need the string version and/or a separate directory with multiple botparsers
* The new call in the construct looks like it breaks the current standard, but creating an object here isn't much different from calling the function with a string and creating that object in the function. As we don't need the string-method for the bot, it seems best to just call it like this.


NanneHuiges:
I wasn't sure if it was better to start a discussion (issue?) first before just doing a pull-request for this, but as the amount of code is small and most is based upon code that's in different equivalent functions, this seemed to be the best way :)

DeviceDetector.php
```diff
@@ -590,14 +611,18 @@ protected function parseBot()
             return false;
         }
 
-        $botParser = new Bot();
-        $botParser->setUserAgent($this->getUserAgent());
-        $botParser->setYamlParser($this->getYamlParser());
-        $botParser->setCache($this->getCache());
-        if ($this->discardBotInformation) {
-            $botParser->discardDetails();
```
sgiehl: This removes the special `discardDetails` for the bot parser. This is quite important, as it speeds up the detection for bots a lot if no detailed bot information is needed.

NanneHuiges: Ow, that's really bad, sorry for that! I'll write a fix for that tomorrow, tried to make it too much te same as the other 2 functions, and then I stopped paying attention. 

Are you open for the concept of this MR though?

sgiehl: Sure, why not allowing multiple / custom bot parsers

NanneHuiges: Added a test that picks up this bug (again, that was careless, sorry), and added a fix. Needed to add an abstract class to define the `discardDetails()` method for the bots.

NanneHuiges:
Have you had time to look at the update? Is there anything else I could/should update?

sgiehl:
will have a closer look on Monday

PR #5715, opened 2018-03-05
Add NetSurf browser and engine

j5lx:
NetSurf is an Open Source browser with its own layout engine that has its origins in the RISC OS community but is available for many different platforms.

Since the most obvious short name NS was already taken by Netscape, I went with NE instead. I tried my best to add both the browser and the layout engine properly, but let me know if there’s anything I missed or didn’t do correctly or if another short name would be preferable.


sgiehl:
Looks fine :+1:

Findus23:
I have added the icon in https://github.com/matomo-org/matomo-icons/commit/020b5a86ee322d6db1a8dd23691e7b61e3085a18.

PR #5741, opened 2018-05-25
Python3 port for device-detector

thinkwelltwd:
Thanks for the excellent library. 

The regexes would be very useful in one of my projects so I ported to Python. Hope it can be a useful contribution to your work.


sgiehl:
detection improvements and new ports are always welcome :+1:

PR #5742, opened 2018-05-27
Adds simple autoloader to use DeviceDetector without composer

### LINKED ISSUE #5505: an alternative to "composer"
Hello Stefan this is
not exactly an "issue"
but rather a request to
add installation details for the use of the
library without being obliged to
use "composer".
One could have reasons
(or issues) in not using it

I have found the following post

http://stackoverflow.com/questions/30011230/how-to-use-piwik-device-detector-in-php-project

but I'm not sure it's correct
(also, there is not a spyc.php file
 in my folder)

and I would love some "official" 
advice on your part

thank you and kind regards
Thomas

sgiehl:
fixes #5505

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,28 @@ if (!is_null($result)) {
 
 ```
 
+## Using without composer
+
+Alternatively to using composer you can also use the included `autoload.php`.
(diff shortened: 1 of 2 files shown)


chrisblakley:
Thanks for this- much appreciated!

PR #5747, opened 2018-06-06
Added some Android devices

GregOriol:


regexes/device/mobiles.yml
```diff
@@ -5100,19 +5139,19 @@ Samsung:
...
-    - regex: '(?:SAMSUNG-)?SM-G7105'
+    - regex: '(?:SAMSUNG-)?SM-G710[25]'
       model: 'GALAXY Grand 2'
       device: 'phablet'
     - regex: '(?:SAMSUNG-)?SM-G53(?:0F|0FZ|0Y|0H|0FZ|1F|1H)'
       model: 'GALAXY Grand Prime'
       device: 'phablet'
     - regex: '(?:SAMSUNG-)?SM-G532F'
-      model: 'GALAXY Grand Prime Plus'
+      model: 'GALAXY J2 Prime/Grand Prime Plus'
```
sgiehl: We should decide to use only one name here. Maybe the more familiar / common name if possible.

regexes/device/mobiles.yml
```diff
@@ -5066,10 +5099,16 @@ Samsung:
       model: 'GALAXY J7+ (C7)'
     - regex: '(?:SAMSUNG-)?SM-G610[FMY]'
       model: 'GALAXY J7 Prime'
+    - regex: '(?:SAMSUNG-)?SM-G615[F]'
+      model: 'GALAXY J7 Max'
     - regex: '(?:SAMSUNG-)?SM-J730(?:F|G|GM)'
       model: 'GALAXY J7 Pro'
+    - regex: '(?:SAMSUNG-)?SM-J701(?:F|M)'
+      model: 'GALAXY J7 Nxt/Neo/Core'
```
sgiehl: would use one of the names only here as well

regexes/device/mobiles.yml
GregOriol: I don't disagree, but I could not decide as both "Grand Prime" and "J2" exist as other variants, and I don't know if one of these two is more popular.

regexes/device/mobiles.yml
GregOriol: Same as the other comment: I don't disagree, but don't know which one could be best...
The names for this one seem to be different by region:
* Nxt in India
* Neo in LATAM
* Core in Philippines

GregOriol:
@sgiehl So, what would you suggest regarding your comments?

sgiehl:
@GregOriol  sorry for the delay. To be honest I'm not sure how to handle that best. But using multiple names might be a bit long. Maybe we could use the model name having most google results?

GregOriol:
Names simplified using Google results, I left the alternatives as comments in the yaml, so we can remember/find them later if trends change.

regexes/device/mobiles.yml
```diff
@@ -3407,6 +3417,18 @@ LG:
...
       model: 'V20'
+    - regex: '(?:LG-)?F800(?:S|K|L)?'
+      model: 'V20'
+    - regex: '(?:LG-)?H91[058]'
+      model: 'V20'
+    - regex: '(?:LG-)?H990(?:DS|N|T)?'
+      model: 'V20'
+    - regex: '(?:LG-)?LS997'
+      model: 'V20'
+    - regex: '(?:LG-)?VS995'
```
sgiehl: this one seems to be a duplication.

GregOriol: Do you prefer to regroup them into a single regex?

sgiehl: that's not necessarily needed. but this specific regex is already done some lines before

PR #5748, opened 2018-06-09
Added Mobiistar Android smartphones

GregOriol:


GregOriol:
I don't know how you choose brand codes, I added it as "M6" because most others were already used (MI/MB/MS/...). Let me know if you think some other code would be better!

sgiehl:
The code is fine. Thanks for the PR

Findus23:
@GregOriol Do you have any idea for an square icon?



This can't really be cropped in a proper way.

GregOriol:
@Findus23 maybe use the double "i" with stars?

Findus23:
@GregOriol 
Thanks for the idea. Thankfully the image was large enough that I could crop it.

https://github.com/matomo-org/matomo-icons/commit/c37c3dfd643d4202897d3bf47d469251e4062eba

Issue #5754, opened 2018-07-05
KaiOS seen as Android


liviuconcioiu:
It seems that KaiOS is seen as Android and also no versions detected. Is based on Android, because it's funded by Google, according to this [Wikipedia](https://en.wikipedia.org/wiki/KaiOS) article.

`Mozilla/5.0 (Mobile; LYF/F30C/LYF_F30C-000-09-10-140318; Android; rv:48.0) Gecko/48.0 Firefox/48.0 KAIOS/2.0`

liviuconcioiu:
[Cunaguaro](https://www.ecured.cu/Cunaguaro_(navegador_web)) browser name, not detected. Is based on Firefox, and is made by a Venezuelan university.

`Mozilla/5.0 (X11; Linux x86_64; rv:27.0) Gecko/20100101 Firefox/27.0 Cunaguaro/27.0`

PR #5757, opened 2018-07-06
Adds detection for KaiOS

sgiehl:
refs #5754


Findus23:
Added icon in https://github.com/matomo-org/matomo-icons/commit/209e4f5fdc37356bb6341c15faa7373fe636fa01

PR #5766, opened 2018-07-15
Added a few Android devices

GregOriol:


regexes/device/mobiles.yml
```diff
@@ -5206,13 +5260,21 @@ Samsung:
...
+    - regex: '(?:SAMSUNG-)?SM-A910F'
+      model: 'GALAXY A9 Pro'
     - regex: '(?:SAMSUNG-)?GT-I9301I'
       model: 'GALAXY S III Neo'
     - regex: '(?:SAMSUNG-)?SM-J120[AFHMT]'
       model: 'GALAXY J1'
-    - regex: '(?:SAMSUNG-)?SM-J111[FGLM]'
+    - regex: '(?:SAMSUNG-)?SM-J110[FHML]'
+      model: 'GALAXY J1 Ace'
+    - regex: '(?:SAMSUNG-)?SM-J111F'
```
sgiehl: Did you remove `J111[GLM]` on purpose?

regexes/device/mobiles.yml
```diff
@@ -910,6 +921,12 @@ Asus:
     - regex: 'Slider SL101'
       model: 'Eee Pad Slider SL101'
       device: 'tablet'
+    - regex: 'P008 Build[);/ ]'
+      model: 'ZenPad Z8'
+      device: 'tablet'
+    - regex: 'P027 Build[);/ ]'
```
sgiehl: having the `[);/ ]` here is not really needed, as it is already matched in the overall regex and `Pxxx Build` should be enough for an exact match then

sgiehl (changes requested):

regexes/device/mobiles.yml
GregOriol: Indeed, changed!

regexes/device/mobiles.yml
GregOriol: I don't think they actually exist: couldn't find information on those on the internet, and there were no tests/user-agents on them.

sgiehl: There seems to be at least a `SM-J111M` (see https://www.handsetdetection.com/device-detection-database/devices/samsung/sm-j111m/)

GregOriol: Ok, indeed, added this one back!

PR #5783, opened 2018-08-24
Add support for Android webview.

richgerdes:
Android WebView is the default engine which comes installed on all android devices. Its used by the Android for system level interfaces, and for displays within apps. The platform is similar but not the same as Chrome for Android (Chrome Mobile), and has limitations around the actions which a developer may perform (such as breaking out of the window, and executing other interfaces.)


sgiehl:
those should meanwhile be detected as chrome webview

PR #5789, opened 2018-08-28
Providing detection of iOS 10 and 11

wronan:
Sorry for the wildcards in minor version checks - I wasn't able to figure out specific versions. Still, better then nothing :)

### DIFF
--- a/regexes/oss.yml
+++ b/regexes/oss.yml
@@ -273,6 +273,38 @@
 ##########
 # iOS
 ##########
+- regex: 'CFNetwork/889\.\d\.\d'
+  name: 'iOS'
+  version: '11.1'
+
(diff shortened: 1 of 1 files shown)


wronan:
Oh man... sorry. I've checked some random commits updating the regexes and haven't found any tests - so I thought that you don't test all minor rules updates like this one. I may update the tests if you like.
Anyway, of you want to check it by yourself here you have a user-agent for iOS 10.3
```
SomeRandomApp/32123 CFNetwork/811.5.4 Darwin/16.7.0
```

sgiehl:
@wronan  Thanks again for your approach. I've copied and improved your changes to a new branch, where I've also added some tests. Will create a PR soon and merge it.

Issue #5792, opened 2018-08-31
Conflicting regexes for Instacast


etienne-martin:
I found an issue with the way we detect Instacast. The regexes are conflicting.

As you can see the following regexes are very similar:

**Feed readers:** 
`Instacast/(\d+) CFNetwork/([\d\.]+)`

**Media players:**
`Instacast(?:/([\d\.]+))? CFNetwork/([\d\.]+)`

**Mobile apps:**
`Instacast(?:HD)?/(\d\.[\d\.abc]+) CFNetwork/([\d\.]+) Darwin/([\d\.]+)`

For example, `Instacast/2380 CFNetwork/720.2.4 Darwin/14.1.0 (x86_64)` gets detected as a `feed reader` while it's a `media player`.

There seems to be an issue with the way the tests are executed because it's not getting caught by the tests.

sgiehl:
Interesting. Didn't know that is in multiple files... 
What exactly is Instacast? We should remove the the detections that are not correct...

Findus23:
@sgiehl It's a podcast app, so in a way a mobile app, a feed reader and a media player at the same time.

etienne-martin:
The Instacast userAgents are pretty much the same between the mobile app, feed reader and media player. We'll have a hard time differentiating between these clients. 

**Feed reader:**
`Instacast/2380 CFNetwork/673.3 Darwin/13.4.0 (x86_64) (MacBookPro5%2C4)`

**Media player:**
`Instacast/2380 CFNetwork/720.2.4 Darwin/14.1.0 (x86_64)`
`Instacast/4.1.2 CFNetwork/672.0.8 Darwin/14.0.0`

**Mobile app:**
`Instacast/5.0a8 CFNetwork/711.1.11 Darwin/13.4.0`
`InstacastHD/1.1.2 CFNetwork/711.0.6 Darwin/14.0.0`

Also, looks like Instacast got discontinued and replaced with the Instacast Core app for iOS according to this article: https://martinhering.me/post/vemedio-product-development-has-been-discontinued

I propose that we only detect it as a mobile app from now on.

PR #5793, opened 2018-09-01
Detect Instacast only as a mobile app.

etienne-martin:
Related to this issue: https://github.com/matomo-org/device-detector/issues/5792


etienne-martin:
I also added an additional test not related to the issue mentioned above.

I discovered that line 692 of `DeviceDetector.php` (https://github.com/matomo-org/device-detector/blob/master/DeviceDetector.php#L692) was not covered by any tests.

PR #5804, opened 2018-09-12
Add Samsung Galaxy NOTE 9 (SM-N960)

pongee:
Source: 
https://www.phonemore.com/search/?q=Samsung%20Galaxy%20Note%209


sgiehl:
Would you mind adding a test with a useragent of that device?

pongee:
Hi. Of course, I added a test.

PR #5822, opened 2018-10-05
Add Oppo smartphones: PACM00, and PACT00

pongee:
Source:
https://www.handsetdetection.com/device-detection-database/devices/oppo/pacm00/
https://www.handsetdetection.com/device-detection-database/devices/oppo/pact00/


regexes/device/mobiles.yml
```diff
@@ -4900,6 +4900,8 @@ OPPO:
       model: 'F1s'
     - regex: '(?:OPPO[ _]?)?R1011 Build'
       model: 'Joy Plus'
+    - regex: '(?:OPPO[ _]?)?PAC[T|M]00 Build'
```
mimmi20: this matches `PAC|00 Build` too

sgiehl: good catch, will change that

PR #5825, opened 2018-10-08
Detect devices MTC, Primepad, Lemnov

sanchezzzhak:
MTC devices
Primepad devices
Lemnov devices
Asus devices
Texet devices


sanchezzzhak:
MTC - Russian operator creates its """masterpieces""" on the basis of  ZTE

regexes/device/mobiles.yml
```diff
@@ -3013,7 +3013,7 @@ IconBIT:
       model: 'NetTAB Thor IZ'
     - regex: 'NT-1009T'
       model: 'NetTAB Thor quad II'
-    - regex: 'NT-1011T'
+    - regex: 'NT-1011T|NT-1001T'
```
sgiehl: could by simplified to `'NT-10[01]1T'`

regexes/device/mobiles.yml
```diff
@@ -3882,6 +3926,8 @@ LG:
       model: '$1'
     - regex: '(VX[0-9]+)'
       model: '$1'
+    - regex: '(Nexsus 5)'
```
sgiehl: Isn't that the same as Google's `Nexus 5` (which was produced by LG)

regexes/device/mobiles.yml
```diff
@@ -7527,6 +7586,32 @@ Zen:
...
+Lemhoov:
+  regex: 'Lemhoov'
+  device: 'smartphone'
+  models:
+    - regex: 'Lemhoov[ _-]([^/;]+) Build'
+      model: '$1'
+
+#Libra
+Libra:
+  regex: 'Libra[ _-]'
```
sgiehl: Is the `Libra 97` maybe a product of the brand `GoClever`. Wasn't able to find it in another context...

regexes/device/mobiles.yml
```diff
@@ -7527,6 +7586,32 @@ Zen:
...
+  models:
+    - regex: 'Libra[ _-](97) Build'
+      model: '$1'
+
+#MTC
+MTC:
+  regex: 'MTC[ _-](978|982|982O|982T)|SMART[ _]Race[ _]4G Build'
+  device: 'smartphone'
+  models:
+    - regex: 'MTC[ _-](978|982|982O|982T) Build'
```
sgiehl: Guess the ` Build` here isn't required, as the global regex doesn't contain it as well.

sgiehl (changes requested):
Thanks for the PR!
Left a few minor comments

regexes/device/mobiles.yml
sanchezzzhak: I'm not really sure it's Google.
Nexsus 5 China DIY device
Nexus 5  Google+LG

sgiehl: Ah so it's a copy of the Nexus 5?

PR #5832, opened 2018-10-10
Improves detection of various Alcatel devices

liviuconcioiu:
refs #5808 - improves detection of various Alcatel devices


sgiehl:
For new devices it would be good to add some tests to ensure the detections work as expected. 
Also I think it doesn't make sense to add `Single SIM` to the name as that is kind of default.

liviuconcioiu:
I've added the tests and removed Single SIM, but technology is failing on me. I can't get them up at the same time. Can you help? Thanks!

sgiehl:
closing in favor of #5834

PR #5834, opened 2018-10-10
Improves detection of various Alcatel devices

liviuconcioiu:
refs #5808 - improves detection of various Alcatel devices


sgiehl:
will merge that in another branch to fix & update the tests before merging

PR #5850, opened 2018-10-20
Adds detection for ComTrade Tesla devices

liviuconcioiu:
refs #5808 - adds detection for ComTrade Tesla devices


Parser/Device/DeviceParserAbstract.php
```diff
@@ -127,6 +127,7 @@
         'CU' => 'Cube',
         'CY' => 'Coby Kyros',
         'C6' => 'Comio',
+		'C7' => 'ComTrade Tesla',
```
sgiehl: please use spaces instead of tabs here

sgiehl:
Any reason for naming the brand `ComTrade Tesla`? Doesn't the brand is called `Tesla`, even if it's currently owned by `Comtrade`.

liviuconcioiu:
I tried to do that, but it conflicts with the Tesla car browser...

sgiehl:
Ah forget about that. You're right... So guess it's fine to name it that way.

PR #5852, opened 2018-10-21
Improves detection of various Bmobile devices

liviuconcioiu:
Improves detection of various Bmobile devices


regexes/device/mobiles.yml
```diff
@@ -1407,11 +1407,17 @@ Blaupunkt:
...
-    - regex: 'AX1075'
-      model: 'AX1075'
+    - regex: 'AX10([1234567]0|[1345678]5|16|65E|70e|73\+|7[1234]|91)[);/ ]'
+      model: 'AX10$1'
+    - regex: 'AX(5[12347]0|512|5[123]5|524|6[01256789]0|[67]05|675|68[0135]|680\+|710|745|810|82[013]|92[012])[);/ ]'
+      model: 'AX$1'
+    - regex: 'AX-340'
+      model: 'AX340'
+    - regex: 'AX7OO'
+      model: 'AX700'
```
sgiehl: Are there any devices of other brands starting with AX? Otherwise we maybe could simplify the whole detections by using `AX-?([1-9][0-9]{2,3})` for all and in the overall match

liviuconcioiu:
From what I've seen there is only one, `Mozilla/5.0 (Linux; Android 4.4.2; AX4Nano Build/KOT49H) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/30.0.0.0 Mobile Safari/537.36`, Allview AX4 Nano.

sgiehl:
ok. but that wouldn't match the regex I mentioned before

regexes/device/mobiles.yml
```diff
@@ -1407,17 +1407,11 @@ Blaupunkt:
 
 # Bmobile
 Bmobile:
-  regex: 'Bmobile[_ ]|AX10(?:[1234567]0|[1345678]5|16|65E|70e|73\+|7[1234]|91)|AX(?:5[12347]0|512|5[123]5|524|6[01256789]0|[67]05|675|68[0135]|680\+|710|745|810|82[013]|92[012])|AX-340|AX7OO'
+  regex: 'Bmobile[_ ]|AX-?([1-9][0-9eEO+]{2,4})'
```
sgiehl: I guess the `eEO+` can only be at the end, right? So maybe we should use `AX-?([1-9][0-9]{2,3}[eEO+])`

liviuconcioiu: You forgot to add a `?` sign. Also, they made a typo mistake on the user agent, instead of 700 they wrote 7OO, but on their website the model is 700. I think this is the final regex `AX-?([1-9][0-9]{2,3}[eEO+]?|7OO)`

PR #5863, opened 2018-10-26
Add 'Baidu Box APP' regex

hiscaler:


sgiehl:
Thanks for the PR. would you mind updating the tests, so the pass again?

PR #5871, opened 2018-11-19
Added Breaker to client regular expressions.

scrothers:
Breaker is a social podcasting application, it is known to scrape RSS files to update their own internal records. Their full user agent is actually:
Breaker/v315 (subscribers=9999; feed-id=123456; url=https://www.breaker.audio/url-slug-to-podcast)

However, there are no concessions at the moment to parse out the additional information in DeviceDetector (yet?).


sgiehl:
Thanks for the PR. Would you mind adding a test with the example useragent you provided?

PR #5876, opened 2018-11-29
Add Pecl YAML parser

mallsbill:


mimmi20:
@mallsbill Please add some information about this optional dependency in the composer.json.

mallsbill:
> @mallsbill Please add some information about this optional dependency in the composer.json.

@mimmi20 I added ext-yaml to suggest

Last travis job failed on "composer require doctrine/cache ~1.2" on PHP 5.3 build
But it should work if you re-run it

Yaml/Pecl.php
```diff
@@ -0,0 +1,13 @@
+<?php
+namespace DeviceDetector\Yaml;
+
+class Pecl implements Parser
```
sgiehl: Would you mind adding a docblock for the class describing the purpose?

Yaml/Pecl.php
```diff
@@ -0,0 +1,13 @@
...
+{
+    public function parseFile($file)
+    {
+        if(function_exists(yaml_parse_file) === false)
+            throw new \Exception('Pecl YAML extension is not installed');
```
sgiehl: please add brackets even for one line if statements.

sgiehl (changes requested):
Left some comments, otherwise looks good :+1:

sgiehl (approved):

PR #5895, opened 2019-01-27
Add tenta browser detection

tufanbarisyildirim:
Tenta is your[ private, encrypted browser](https://tenta.com/) that protects your data instead of selling. We're building a next-generation browser that combines all the privacy tools you need, including built-in OpenVPN. Everything is encrypted by default. That means your bookmarks, saved tabs, web history, web traffic, downloaded files, IP address and DNS. A truly incognito browser that's fast and easy.
We as Tenta are also fan of piwik and we use it for analytics for our projects. We added our own browser to library to see more accurate statistics.


Parser/Client/Browser.php
```diff
@@ -211,6 +212,7 @@ class Browser extends ClientParserAbstract
         'Opera'              => array('OP', 'OM', 'OI', 'ON', 'OO'),
         'Safari'             => array('SF', 'MF'),
         'Sailfish Browser'   => array('SA'),
+        'Tenta Browser'      => array('TB')
```
Findus23: I'm not sure if this is a separate browser family as I assume this is based on the chromium android source, right?

Findus23 (commented):
Apart from one note this change is looking good.
BTW: Have you thought about open sourcing the browser? I'd expect a large portion of your target group to never use a browser where they can't know for sure what it does exactly.

Parser/Client/Browser.php
tufanbarisyildirim: Please check our [github account](https://github.com/tenta-browser) that you will find most of the engine and security components are opensourced. We are opensourcing every component that is ready to be public. So, not for all parts but the goal is to be fully opensourced. Also about being a browser family, we are beasd on Chromium with our own encrypted communication layer. So, yeah I didn't understand the family part at first look. Will change it and push again. Thanks for feedback @Findus23

README.md
```diff
@@ -192,19 +192,17 @@ There are already a few ports of this tool to other languages:
 - **Crystal** https://github.com/creadone/device_detector
 
 
-## What Device Detector is able to detect
-
```
sgiehl: could you please revert this deletion?

sgiehl (changes requested):

README.md
tufanbarisyildirim: ups! done

sgiehl (approved):

PR #5911, opened 2019-02-11
Improves detection for various devices

liviuconcioiu:
Improves detection of AllCall, AllView, CUBOT, Evolio, Google, Huawei, Karbonn, M4tel, NYX Mobile, Nokia, Oppo, Sharp, Sony, Timovi, ZTE devices

### DIFF
--- a/regexes/device/mobiles.yml
+++ b/regexes/device/mobiles.yml
@@ -7181,7 +7181,7 @@ Spice:
 
 # Sharp
 Sharp:
-  regex: 'SHARP|SBM|SH-?[0-9]+[a-z]?[);/ ]|AQUOS|506SH|SHL22|FS8010 '
+  regex: 'SHARP|SBM|SH-?[0-9]+[a-z]?[);/ ]|AQUOS|506SH|SHL22|FS8010[);/ ]'
   device: 'smartphone'
   models:
     # explicit smartphone models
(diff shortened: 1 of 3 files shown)


regexes/device/mobiles.yml
```diff
@@ -7057,7 +7181,7 @@ Spice:
 
 # Sharp
 Sharp:
-  regex: 'SHARP|SBM|SH-?[0-9]+[a-z]?[);/ ]|AQUOS|506SH|SHL22 '
+  regex: 'SHARP|SBM|SH-?[0-9]+[a-z]?[);/ ]|AQUOS|506SH|SHL22|FS8010 '
```
sgiehl: Actually I'm nor sure if the ` ` in the end of `SHL22` was there on purpose. But as there are no tests that fail I guess it wasn't. Nevertheless I think it might be better to match exact model fragments like `SHL22` or `FS8010` with a `[);/ ]` in the end, so it doesn't match anything with additional chars

regexes/device/mobiles.yml
```diff
@@ -7399,6 +7525,10 @@ Timovi:
   regex: 'Tmovi'
   device: 'smartphone'
   models:
+    - regex: 'Prime'
```
sgiehl: Might be better to use `Tmovi[ _]Prime`, so it also matches `Mozilla/5.0 (Linux; U; Android 4.4.2; es-LA; Tmovi_Prime Build/KOT49H) AppleWebKit/528.5+ (KHTML, like Gecko) Version/3.1.2 Mobile Safari/525.20.1 UCBrowser/9.7.0.520 Mobile`

sgiehl (changes requested):

sgiehl (approved):

PR #5912, opened 2019-02-11
Adds detection for 3 new browsers

liviuconcioiu:
Adds detection for AOL Shield, TenFourFox and Qwant Mobile browsers


sgiehl (changes requested):
Guess we should add the new browsers to the browser family arrays. Qwant Mobile and TenFourFox seem to be FireFox and AOLShield is Chrome, right?

sgiehl (approved):

PR #5914, opened 2019-02-11
Prevent conflicts from other classes autoloader functions

chrisblakley:
I am using Device Detector within WordPress, and while I love the new built-in autoloader.php functionality, I noticed that some other core and plugin classes (like WP_Filesystem and Woocommerce) are conflicting with this autoloader and triggering `No such file or directory` PHP warnings.

By checking the `$class` just inside of the `dd_autoload` function for `DeviceDetector\` with this conditional: `strpos($class, 'DeviceDetector\\') !== false` this problem is resolved.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -14,17 +14,19 @@
  */
 function dd_autoload($class)
 {
-    $namespace_map = array('DeviceDetector\\' => __DIR__ . '/');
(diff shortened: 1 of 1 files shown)


sgiehl:
makes sense 👍

Issue #5934, opened 2019-03-08
Wrong description for type of $device


zaycker:
```
/**
 * Holds the device type after parsing the UA
 * @var string
 */
protected $device = null;
```

but

```
/**
* Returns the device type extracted from the parsed UA
*
* @see DeviceParserAbstract::$deviceTypes for available device types
*
* @return int|null
*/
public function getDevice()
{
    return $this->device;
}
```

zaycker:
found this inconsistency in a process of migration to Go

sgiehl:
fixed in #5935

PR #5935, opened 2019-03-08
Wrong description for type of $device (#5934)

zaycker:
increased performance by x2


sgiehl:
doesn't make a difference for PHP, but good to fix nevertheless ;-)

PR #5936, opened 2019-03-08
Improves detection for Samsung devices

liviuconcioiu:
Improves detection for Samsung devices


sgiehl (changes requested):
Sorry for the delay. Left some suggestions

Issue #5959, opened 2019-05-10
Lowercase in useragent vs generic matching


GregOriol:
I have found this user agent that has the device name in lowercase:
```
Mozilla/5.0 (Linux; Android 9; moto g(6) plus) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.90 Mobile Safari/537.36
```

The matched rule is (simplified):
```
- regex: 'Moto ([CGEZ]))'
model: 'Moto $1'
```

The problem is that this marks the model as "Moto g" instead of "Moto G".

We could make a specific rule to match "Moto G" and set the model to "Moto G" but that seems a bit extra work since the generic regex is nice.

What do you think? What would be a good way to handle that situation?

PR #5960, opened 2019-05-10
Enhanced replacement options for generic regexes

GregOriol:
As a followup for #5959, I've tried an idea and would like to suggest this change to have enhanced replacement options for generic regexes.

The idea is that in addition to $1 $2 $3 remplacements, we could also have modifiers like $u1 $l2 $uf3 to replace by an uppercase lowercase or uppercase-first value.

This would allow to keep the current generic non-case-sensitive regexes, while also supporting some user-agents where parts are in a wrong/outrageous case, like the example in the fixture with "Moto G" being provided as "g" in the user-agent.

This is backwards compatible and doesn't slow down the detection.

### DIFF
--- a/Parser/ParserAbstract.php
+++ b/Parser/ParserAbstract.php
@@ -194,13 +194,14 @@ abstract class ParserAbstract
     protected function buildByMatch($item, $matches)
     {
         for ($nb=1;$nb<=3;$nb++) {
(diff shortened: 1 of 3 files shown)


sgiehl:
nice improvement. Guess we could merge for the next major release. Can't merge that right away, as it might break some of the ports to other languages reusing the yml files.

PR #5961, opened 2019-05-10
Added a few Android devices

GregOriol:
NB: this PR could be enhanced by PR #5960 as the following commits have lowercase model names and they should be uppercased:
* 01d8853
* be9f4f1
* e97c44a


regexes/device/mobiles.yml
```diff
@@ -6164,14 +6166,8 @@ Neffos:
...
   models:
-    - regex: 'Neffos ([^/;]+) Build'
-      model: '$1'
-    - regex: 'Neffos ([^/;]+)/'
-      model: '$1'
-    - regex: 'Neffos_C9A'
-      model: 'C9A'
-    - regex: 'Neffos_(C|X)9'
-      model: '$19'
+    - regex: 'Neffos[_ ]((?:C|X)[0-9][A-Z]?(?: Max)?)'
```
sgiehl: Any reason for removing the generic regexes? Useragents like `Mozilla/5.0 (Linux; Android 7.0; Neffos X1 Lite Build/NRD90M) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.84 Mobile Safari/537.36` won't be detected correctly with the new one.

Also the one you used could be simplified to:
`Neffos[_ ]([CX][0-9][A-Z]?(?: Max)?)` and `model` should only be `$1` as `$2` won't be set ever.

sgiehl (changes requested):

GregOriol:
@sgiehl Indeed, the model for Neffos has an extra $2, it's something I forgot to remove after changing the regexes...

The new regex is actually set up so that it matches all the same devices as the previous ones, and it is even more generic and future proof if more variants appear. All the tests pass, so all known user agents work with it. I'm adding yours in a new commit.

regexes/device/mobiles.yml
GregOriol: @sgiehl Indeed, the model for Neffos has an extra $2, it's something I forgot to remove after changing the regexes...

The new regex is actually set up so that it matches all the same devices as the previous ones, and it is even more generic and future proof if more variants appear. All the tests pass, so all known user agents work with it. I'm adding yours in a new commit.

sgiehl (approved):

PR #5963, opened 2019-05-18
Improves detection for various devices

liviuconcioiu:
Improves detection of Huawei, Samsung devices


regexes/device/mobiles.yml
```diff
@@ -3843,6 +3853,8 @@ Huawei:
       model: 'P8 Lite (2015)'
     - regex: 'PRA-L(?:A1|X2|X1|X3)|hi6210sft'
       model: 'P8 Lite (2017)'
+    - regex: 'SNE-AL00'
+      model: 'Maimang 7'
```
sgiehl: Maybe we should use the alias `Mate 20 Lite` here, as we didn't use `Maimang` for other devices before...

sgiehl (changes requested):

PR #5975, opened 2019-06-18
Detect devices: Odys, Advan, Evercoss, Dapeng, Figo

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -2380,6 +2410,14 @@ Datsun:
     - regex: 'DATSUN_D5500'
       model: 'D5500'
 
+# Dapeng
+Dapeng:
+  regex: 'A75 Build'
```
sgiehl: Not sure if that detection doesn't match to many things. Actually there are other devices named A75, like https://www.oppo.com/en/smartphone-a75/ or one from Micromax. Guess we shouldn't add it if it isn't unambiguous?

regexes/device/mobiles.yml
```diff
@@ -7141,12 +7232,22 @@ Roku:
...
+  regex: 'IO Pro'
+  device: 'smartphone'
+  models:
+    - regex: 'IO Pro'
+      model: 'IO Pro'
+
 # Rover
 Rover:
-  regex: 'Rover ([0-9]+)'
+  regex: 'rokit io pro 3d'
```
sgiehl: this change does not seem to make any sense, guess you wanted to add that in the detection before...

sgiehl (changes requested):

regexes/device/mobiles.yml
sanchezzzhak: Remove.

regexes/device/mobiles.yml
sanchezzzhak: Revert change

PR #5984, opened 2019-07-03
Detect QMobile, Fondi, ANS, STF

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -1261,6 +1261,14 @@ Ask:
     - regex: 'SP618'
       model: 'SP618'
 
+# ANS
+ANS:
+  regex: 'UL40 Build'
+  device: 'smartphone'
+  models:
+    - regex: '(UL40)'
+      model: '$1'
```
sgiehl: that could be changed to a static detection, eg
```
    - regex: 'UL40'
      model: 'UL40'
```

regexes/device/mobiles.yml
```diff
@@ -8666,6 +8693,12 @@ Symphony:
 
 # Qtek
 Qtek:
+  regex: '(Q09) Build'
+  device: 'smartphone'
+  model: '$1'
+
+# Q-Touch
+Q-Touch:
```
sgiehl: guess that's at an incorrect position. the new regex for `Q09` should be below `Q-Touch`. You moved the `Qtek` detections to the new one...

sgiehl (changes requested):

Issue #5980, opened 2019-06-30
support more firefox browsers


Findus23:
It seems like recently the number of firefox browsers has increase by quite a bit, so we should maybe add some of them to devicedetector:

### Firefox Lite/ Firefox Rocket  ✔️ 
https://play.google.com/store/apps/details?id=org.mozilla.rocket
https://github.com/mozilla-tw/FirefoxLite

It seems to be quite similar to Firefox Klar/Focus, but focus more on saving data by not loading ads, images, etc. and working on very low resource devices.
It's based on webkit/webviews and developed by Mozilla Asia.
The user agent seems to be quite helpful:
`Mozilla/5.0 (Linux; Android 9; ONEPLUS A3003 Build/PQ3A.190505.002; rv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Rocket/1.6.2(12322) Chrome/74.0.3729.157 Mobile Safari/537.36`

### Firefox Preview / Fenix
https://play.google.com/store/apps/details?id=org.mozilla.fenix

This is the main proof of concept that should soon replace Firefox on Android and become the main browser. It's based on Gecko Android Components with try to be similar to webviews, but with Mozillas browser engine.

Unfortunatly the user agent I get on my phone is not all that helpful:
`Mozilla/5.0 (Android 9; Mobile; rv:68.0) Gecko/68.0 Firefox/68.0`

If someone has some ideas on how to differentiate these browsers, I'm all for it.

sgiehl:
As Fenix seems not to be detectable, I'll close this issue for now. If the useragent changes sometimes and it would be possible to detect, feel free to create a new issue.

PR #5992, opened 2019-07-22
#5980 Detect Browser Firefox Rocket

sanchezzzhak:


Parser/Client/Browser.php
```diff
@@ -206,7 +207,7 @@ class Browser extends ClientParserAbstract
         'BlackBerry Browser' => array('BB'),
         'Baidu'              => array('BD', 'BS'),
         'Amiga'              => array('AV', 'AW'),
-        'Chrome'             => array('CH', 'BA', 'BR', 'CC', 'CD', 'CM', 'CI', 'CF', 'CN', 'CR', 'CP', 'IR', 'RM', 'AO', 'TS', 'VI', 'PT', 'AS', 'TB', 'AD', 'SB'),
+        'Chrome'             => array('CH', 'BA', 'BR', 'CC', 'CD', 'CM', 'CI', 'CF', 'CN', 'CR', 'CP', 'IR', 'RM', 'AO', 'TS', 'VI', 'PT', 'AS', 'TB', 'AD', 'SB', 'FR'),
```
sgiehl: Is Firefox Rocket really based on Chrome?

sanchezzzhak: I removed the binding to chrome
a quick inspection didn't show me the chromium. https://github.com/mozilla-tw/FirefoxLite

Findus23:
As I explained in https://github.com/matomo-org/device-detector/issues/5980, Firefox Rocket is essentially a webview using the Android system Webkit renderer. So it doesn't use Chromium source code but essentially it is using Chromium to render the page.
I wouldn't add it to the chrome-base list, but to the webkit one.

sgiehl (approved):
LGTM. 

@Findus23  should we add it to the Firefox browser family? The rendering engine is detected separately.

Findus23:
Hm, it makes sense for people to expect this category as "Browsers Mozilla develops under the Firefox brand" and therefore it should be in the Firefox family.

PR #6005, opened 2019-07-31
Adds detection of RSSRadio podcast client and RSSRadio Bot

poschi3:
I have RSSRadio in my podcast website logs.

As a [iOS only client](http://rssrad.io/#/rssradio):
* `RSSRadio/9220 (iPad;iOS;12.3.1)`
* `RSSRadio/9252`

As a Bot
* `RSSRadio (Push Notification Scanner;support@dorada.co.uk)`


regexes/oss.yml
```diff
@@ -425,6 +425,9 @@
   name: 'iOS'
   version: ''
 
+- regex: 'RSSRadio/'
+  name: 'iOS'
+  version: '$1'
```
sgiehl: This one could be merged with the one 2 above: `'Podcasts/(?:[\d\.]+)|Instacast(?:HD)?/(?:\d\.[\d\.abc]+)|Pocket Casts, iOS|Overcast|Castro|Podcat|i[cC]atcher'`. Also the `$1` would be useless here, as the regex can't match anything

poschi3: Good point. I'll try it. Should I rebase and ammend my commits?

sgiehl: no need to rebase or amend. will be squashed when merging

poschi3:
I think the PR is ready again :-)

sgiehl (approved):

PR #6011, opened 2019-08-15
add new browsers, apps and devices

mimmi20:
This PR includes changes from PR #5863 and conflicts with the changes of PR #5783.


sgiehl:
I see what you are trying to do with adding new tests. Could you maybe move adding your new fixtures after the already existing fixtures have been sorted? That way they will simply be added at the end and it's easier to see the actual changes and it won't conflict with other PRs that much.

Also I'm not sure why the surrounding `"` from the version numbers have been removed everywhere.
That actually doesn't hurt for the PHP tests here, but some of the ports to other languages are using the test fixtures here, and they have a problem as they are "detected" as invalid numbers without those surrounding `"`
Guess the script to add those chars didn't work on your system:
https://github.com/matomo-org/device-detector/blob/6bb8507bd06f041b073924368f18f0b7f74b7402/misc/sortTestFixtureFiles.php#L116

mimmi20:
@sgiehl The PR is not finished yet. I work on a windows system, and `sed` doesn't work there. This will be fixed tomorrow. I used the `misc/sortTestFixtureFiles.php` file to rewrite and add the test cases.

mimmi20:
@sgiehl You may review now.

Parser/Client/Browser.php
```diff
@@ -71,15 +73,18 @@ class Browser extends ClientParserAbstract
         'CY' => 'Cyberfox',
         'CS' => 'Cheshire',
         'CU' => 'Cunaguaro',
+        'CV' => 'Chrome Webview',
+        'DA' => 'Dalvik',
```
Findus23: Dalvik isn't a browser, but something android webviews sometimes send as a useragent:
https://stackoverflow.com/questions/23804278/browser-sending-dalvik-as-user-agent

Findus23 (changes requested):
I can't say much about the non-browser changes, but I am not sure if spliting up iOS/mobile browser variants is that useful. After all DeviceDetector is already able to keep them apart via the OS detection.

mimmi20:
I will split the PR.

PR #6013, opened 2019-08-18
sort available browsers by key

mimmi20:
Because the available browsers are not sorted by key nor by name, I sorted them by key. This is easier to find free keys for new browsers.


sgiehl:
guess that would conflict with your other PR. Also as the list of known browsers will be completely removed from the parser in v4, I will close that PR. Thanks anyway for the approach.

PR #6014, opened 2019-08-18
Prepare new major release 4.0

### LINKED ISSUE #5951: Unify empty responses
Changing this would probably break backwards compatibility, so I'm not sure if this could be changed, but I want to document this nevertheless:

The output of the different methods return different values if nothing can be detected (null, empty array and "").

Example script:
```php
<?php
require_once 'vendor/autoload.php';
use DeviceDetector\Parser\Device\DeviceParserAbstract;

$userAgent = "thisissomeunknownuseragent";
$dd = new DeviceDetector($userAgent);
$dd->parse();

$botInfo = $dd->getBot();
$clientInfo = $dd->getClient();
$osInfo = $dd->getOs();
$device = $dd->getDeviceName();
$brand = $dd->getBrandName();
$model = $dd->getModel();

var_dump($botInfo);
var_dump($clientInfo);
var_dump($osInfo);
var_dump($device);
var_dump($brand);
var_dump($model);
```
Output:
```
NULL
NULL
array(0) {
}
string(0) ""
string(0) ""
string(0) ""
```

sgiehl:
* increase requirement to PHP 7.1
* use strict types
* improve code quality using phpstan and phpcs
* removes browser & os short codes


Findus23:
@sgiehl What's the plan for removing short code in Matomo?
While this makes things easier and renaming the icons should be pretty easy, I wonder if this might cause new issues in the future. 
What if in the future an OS or Browser get's renamed (which doesn't seem that unlikely)?

sgiehl:
@Findus23 that needs to be handled with https://github.com/matomo-org/matomo/issues/5413
And I would ignore the case if a browser get's renamed. But we could maybe maintain a list of "alias names" in device detector and group the records in matomo with this info

Alternatively if the client detection won't make it into Matomo soon, we need to maintain the shortcode list there.

sgiehl:
Closing as most parts of it were included in another PR

PR #6018, opened 2019-08-21
Update blink detection

mimmi20:
This PR changes the auto-detection for the Blink engine, as it does not use a `Blink` token. 
The test fixtures were rewriten using the script from PR #6017.

### DIFF
--- a/Tests/Parser/Client/fixtures/browser.yml
+++ b/Tests/Parser/Client/fixtures/browser.yml
@@ -1,50 +1,50 @@
 ---
--
+- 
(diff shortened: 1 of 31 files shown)


sgiehl:
I don't think that change makes much sense. Chrome actually switched the engine from WebKit to Blink in version 28. This is also in the browser definition. See https://github.com/matomo-org/device-detector/blob/master/regexes/client/browsers.yml#L658-L664
If that isn't defined for some Chrome based browsers, that might simply be missing...

PR #6022, opened 2019-08-27
Add "Samsung Note 10" and "Samsung Note 10+" smartphones

pongee:


regexes/device/mobiles.yml
```diff
@@ -8916,6 +8916,9 @@ Samsung:
     - regex: '(?:SAMSUNG[- ])?SM-N960'
       model: 'GALAXY Note 9'
       device: 'phablet'
+    - regex: '(?:SAMSUNG[- ])?SM-N970[F|U|U1|X]'
```
mimmi20: Doesn't this regex make the char `|` a valid one? Shouldn't this writen like this: `regex: '(?:SAMSUNG[- ])?SM-N970(?:F|U|U1|X)'`

pongee: thank you. I fixed it , and I added the note 10+ as well

PR #6028, opened 2019-09-02
add Crosswalk App

mimmi20:
Is `Streamy` really a browser? Or could I move it to the mobile apps?


sgiehl:
I would count it as a browser: https://www.streamybrowser.com/

PR #6038, opened 2019-09-16
add FreshRSS as bot

Kdecherf:


sgiehl:
could you also update the failing test?

Findus23:
As SimplePie is a library and FreshRSS is an application, I think these should be changed to freshrss:

https://github.com/matomo-org/device-detector/blob/d6dbed84ac647766b69143b015f3289db91b0591/Tests/fixtures/bots.yml#L2236-L2261

Kdecherf:
@Findus23 I will update the first test which is failing and I'll let the two others as is

Kdecherf:
@Findus23 @sgiehl it should be good now

PR #6040, opened 2019-09-25
Added device detect for the following brands: Samsung, Sony, Tecno, MYFON, Vivax, Archos, Megafon, Manta, LG, DEXP, Irbis, Orarge, Eurostart

sanchezzzhak:


sanchezzzhak:
wow fast tests passed :+1:

regexes/device/mobiles.yml
```diff
@@ -7835,7 +7863,7 @@ Orange:
     - regex: 'FunTab ([^/;]+) Build'
       model: 'FunTab $1'
       device: 'tablet'
-    - regex: 'Orange[ _-]([^/;]+) Build'
+    - regex: 'Orange[ _-]([^/;]+)( Build|\))'
```
sgiehl: When using `)` as "stop" character, it should be excluded from the match pattern, as it otherwise could match to much.
Would imho be better to use this
```suggestion
    - regex: 'Orange[ _-]([^/;)]+)( Build|\))'
```

regexes/device/mobiles.yml
```diff
@@ -7823,7 +7862,7 @@ Opsson:
 
 # orange
 Orange:
-  regex: 'SPV[ \-]?([a-z0-9]+)|Orange[ _-]([^/;]+) Build|SC/IHD92|FunTab ([^/;]+) Build'
+  regex: 'SPV[ \-]?([a-z0-9]+)|Orange[ _-]([^/;]+)( Build|\))(?<!Tahiti-LS\))|SC/IHD92|FunTab ([^/;]+) Build'
```
mimmi20: Schould the term `(?<!Tahiti-LS\))` not be before the searched word, not behind?

sanchezzzhak: I tried to put in the beginning but it didn't work
success https://regex101.com/r/PaZaPd/1
bad https://regex101.com/r/PaZaPd/2

mimmi20: I would write the regex this way: `Orange[ _-](?!Tahiti-LS\))([^\/;]+)( Build|\))`

sanchezzzhak: Thanks

PR #6054, opened 2019-10-02
update VK-Mobile and LG devices, add Fengxiang devices

mimmi20:


regexes/device/mobiles.yml
```diff
@@ -10378,9 +10384,15 @@ Vitelcom:
...
+# Fengxiang
+Fengxiang:
+  regex: '(vKB011B|vKB004L)'
+  device: 'tablet'
+  model: '$1'
+
 # VK Mobile
 VK Mobile:
-  regex: 'VK(?!share)[\-]?([a-z0-9 ]+)[ ;\)/]'
+  regex: 'VK(?!share)[\-]?([a-zA-Z0-9 ]+)([;\)/ ]| Build)'
```
sgiehl: the `A-Z` should actually not be needed, as all regexes are done case insensitive in https://github.com/matomo-org/device-detector/blob/master/Parser/ParserAbstract.php#L180

mimmi20: `A-Z` removed

PR #6064, opened 2019-10-08
Detect devices for brands: Cube, Digma, TLS, LG, Qilive, teXet, Unimax, Ark, FORME, Tecno

sanchezzzhak:


Parser/ParserAbstract.php
```diff
@@ -170,19 +170,25 @@ protected function getRegexesDirectory()
...
         // only match if useragent begins with given regex or there is no letter before it
         $regex = '/(?:^|[^A-Z0-9\-_]|[^A-Z0-9\-]_|sprd-)(?:' . str_replace('/', '\/', $regex) . ')/i';
-
-        if (preg_match($regex, $this->userAgent, $matches)) {
-            return $matches;
+        try {
+            if (preg_match($regex, $this->userAgent, $matches)) {
+                return $matches;
+            }
+        } catch (\Exception $exception) {
```
mimmi20: When is this Exception thrown?

Parser/ParserAbstract.php
```diff
@@ -170,19 +170,25 @@ protected function getRegexesDirectory()
...
+            throw new \Exception(
```
mimmi20: Is there no better/less generic Exception to be thrown?

Parser/ParserAbstract.php
sanchezzzhak: If the regular expression has incorrect syntax

mimmi20 (approved):

PR #6065, opened 2019-10-11
Detect devices for brands: Apple, 3Q, Acer, Alcatel, Allview, Ainol, AIS, Motorola, Siemens

mimmi20:
This PR adds or updates some devices.


regexes/device/mobiles.yml
```diff
@@ -8381,7 +8526,7 @@ Santin BiTBiZ:
 
 # Siemens
 Siemens:
-  regex: 'SIEMENS|SIE-|portalmmm/2\.0 SI|S55|SL45i'
+  regex: 'SIEMENS|SIE-|portalmmm/2\.0 SI|(?<!Android 4.4.2; )S55|SL45i'
```
sgiehl: I don't think that makes much sense. If S55 might be another device if it's running android, we shouldn't match an exact version number, as it might be possible that the device runs a newer or older vesion as well.

mimmi20: I changed the Acer regex too, the android part here is not needed anymore.

PR #6070, opened 2019-10-19
Add 7 new crawlers and 3 feed readers

Kdecherf:
Added feed reader:
- QuiteRSS

Added feed readers as bots:
- inoreader.com
- theoldreader.com

Added crawlers:
- BoardRader Blog Indexer
- Nuzzel
- eCairn-Grabber
- nbertaupete95
- Twingly Recon
- The Knowledge AI
- Google Favicon

### DIFF
--- a/regexes/bots.yml
+++ b/regexes/bots.yml
@@ -204,6 +204,13 @@
     name: 'Blogtrottr Ltd'
     url: 'https://blogtrottr.com/'
 
+- regex: 'BoardReader Blog Indexer'
+  name: 'BoardReader Blog Indexer'
+  category: 'Crawler'
+  producer:
+    name: 'BoardReader'
+    url: 'http://boardreader.com/'
+
 - regex: 'BountiiBot'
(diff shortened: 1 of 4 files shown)


regexes/bots.yml
```diff
@@ -1005,6 +1035,11 @@
     name: 'QueryEye Inc.'
     url: 'http://queryeye.com'
 
+- regex: 'QuiteRSS'
+  name: 'QuiteRSS'
+  category: 'Feed Reader'
+  url: https://quiterss.org
```
sgiehl: Are QuiteRSS and the other Feed Readers you added here online services that are crawling websites to present their feeds, or are they applications, that directly scrawl feeds on demand? If it's the latter one, we should add them as feed reader client instead

Kdecherf: You're right, I'll update it in the coming days

Kdecherf: @sgiehl I moved QuiteRSS as a feed reader app

PR #6072, opened 2019-10-20
Adds detection for Hawk Turbo Browser, FreeU, Nox Browser, Basilisk, Sputnik Browser,  K.Browser, Oculus Browser, Jio Browser, hola! Browser, Huawei Browser, vivo Browser and Realme Browser browsers

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -249,7 +252,7 @@ class Browser extends ClientParserAbstract
      * @var array
      */
     protected static $mobileOnlyBrowsers = array(
-        '36', 'PU', 'SK', 'MF', 'OI', 'OM', 'DD', 'DB', 'ST', 'BL', 'IV', 'FM', 'C1', 'AL', 'SA', 'SB', 'FR', 'WP'
+        '36', 'PU', 'SK', 'MF', 'OI', 'OM', 'DD', 'DB', 'ST', 'BL', 'IV', 'FM', 'C1', 'AL', 'SA', 'SB', 'FR', 'WP', 'HA'
```
sgiehl: According to http://www.noxbrowser.com/ the Nox Browser is a mobile browser only as it's only available for Android and iOS

liviuconcioiu: Fixed

Parser/Client/Browser.php
```diff
@@ -231,16 +243,17 @@ class Browser extends ClientParserAbstract
...
         'Internet Explorer'  => array('IE', 'IM', 'PS'),
         'Konqueror'          => array('KO'),
         'NetFront'           => array('NF'),
         'NetSurf'            => array('NE'),
         'Nokia Browser'      => array('NB', 'NO', 'NV', 'DO'),
         'Opera'              => array('OP', 'OM', 'OI', 'ON', 'OO', 'OG', 'OH', 'O1'),
         'Safari'             => array('SF', 'MF', 'SO'),
-        'Sailfish Browser'   => array('SA')
+        'Sailfish Browser'   => array('SA'),
+        'Samsung Browser'    => array('OC')
```
sgiehl: According to https://developer.oculus.com/documentation/oculus-browser/latest/concepts/browser-intro/ Oculus Browser is based on Chromium. So guess we should add it to the Chrome family

liviuconcioiu: Fixed. Also, do you think that a new device type should be added for VR devices? Maybe in a future release?

sgiehl: Guess that would make sense to add new device types where useful

PR #6073, opened 2019-10-20
Adds detection for Yandex Browser Lite and improves version detection for Yandex Browser

liviuconcioiu:
Improves detection for Yandex browsers

### DIFF
--- a/regexes/client/browsers.yml
+++ b/regexes/client/browsers.yml
@@ -522,6 +522,13 @@
   engine:
     default: 'Blink'
 
+# Yandex Browser (beta)
+- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(beta\)?'
+  name: 'Yandex Browser (beta)'
+  version: '$1'
+  engine:
+    default: 'Blink'
+
 # Yandex Browser Lite
 - regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(lite\)?'
   name: 'Yandex Browser Lite'
--- a/regexes/client/browsers.yml
+++ b/regexes/client/browsers.yml
@@ -515,20 +515,6 @@
   name: 'Baidu Spark'
   version: '$1'
 
-# Yandex Browser (alpha)
-- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(alpha\)?'
-  name: 'Yandex Browser (alpha)'
-  version: '$1'
-  engine:
-    default: 'Blink'
-
-# Yandex Browser (beta)
-- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(beta\)?'
-  name: 'Yandex Browser (beta)'
-  version: '$1'
-  engine:
-    default: 'Blink'
-
 # Yandex Browser Lite
 - regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(lite\)?'
   name: 'Yandex Browser Lite'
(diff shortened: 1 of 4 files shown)


regexes/client/browsers.yml
```diff
@@ -515,6 +515,27 @@
...
 
+# Yandex Browser (alpha)
+- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(alpha\)?'
+  name: 'Yandex Browser (alpha)'
+  version: '$1'
+  engine:
+    default: 'Blink'
+
+# Yandex Browser (beta)
+- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(beta\)?'
```
mimmi20: Is the last `?` on the right position?

regexes/client/browsers.yml
```diff
@@ -515,6 +515,27 @@
...
+  name: 'Yandex Browser (beta)'
+  version: '$1'
+  engine:
+    default: 'Blink'
+
+# Yandex Browser Lite
+- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(lite\)?'
```
mimmi20: Is the last `?` on the right position?

regexes/client/browsers.yml
```diff
@@ -515,6 +515,27 @@
   name: 'Baidu Spark'
   version: '$1'
 
+# Yandex Browser (alpha)
+- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(alpha\)?'
```
mimmi20: Is the last `?` on the right position?

regexes/client/browsers.yml
liviuconcioiu: Yes, is based on the regex for Yandex Browser.

regexes/client/browsers.yml
liviuconcioiu: Yes, is based on the regex for Yandex Browser.

regexes/client/browsers.yml
liviuconcioiu: Yes, is based on the regex for Yandex Browser.

sgiehl:
Does it make sense to add detection for Alpha and Beta releases as a separate browser?
Imho those are prerelease versions of Yandex browser and no "new" browsers :thinking:

liviuconcioiu:
It depends on how device-detector is used. For analytics point of view, it doesn't matter so much, but if it's used in affiliate marketing for example, it matters. It can differentiate tech guys (as most of the alpha, beta versions are used by them) for example from other users.

regexes/client/browsers.yml
mimmi20: Should the regex not be `'YaBrowser(?:/(\d+[\.\d]*))? \(alpha\)'` ? The last `?` in the Yandox browser regex makes the version part optional.

sgiehl:
We can for sure include `alpha` or `beta` in the version. But imho not in the browser name.
e.g. use something like:

```
regex: YaBrowser(?:/(\d+[\.\d]*)) ?\((alpha|beta)\)?
version: $1 $2
```

That should result in versions like `17.4.0.491.00 alpha`

regexes/client/browsers.yml
sgiehl:
```suggestion
- regex: 'YaBrowser(?:/(\d+[\.\d]*)) \(lite\)'
```

The `?` in the end is indeed kind of useless. Or might it be possible the the user agent only contains `(lite` (without the closing `)`)

sgiehl:
@liviuconcioiu could you apply the suggested changes and resolve the conflict? Would be good to merge afterwards.

PR #6075, opened 2019-10-25
Fixed version truncation so one client doesn't affect the others

Spikes042:
When `ParserAbstract::$maxMinorParts` is changed by any
Client Parser, it is propagated to all Client Parsers. This
happens because static properties defined ONLY in the parent
class shares a common value with the child class.

To set version truncation specifically for a custom ClientParser, we
need to redefine `ParserAbstract::$maxMinorParts` in the child class
and use late static binding to access the property.

Example:
```php
<?php

use DeviceDetector\Parser\Client\ClientParserAbstract;

class MyAppParser extends ClientParserAbstract{
    protected static $maxMinorParts = self::VERSION_TRUNCATION_PATCH;
    protected $fixtureFile = 'my_app.yml';
    protected $parserName = 'myapp';

    protected function getRegexesDirectory(){
        return __DIR__;
    }
}
````

More on late static binding in PHP:
https://www.php.net/manual/en/language.oop5.late-static-bindings.php


sgiehl:
@Spikes042  sorry for the late response. Actually the version truncation was meant to be global. But guess it doesn't matter to make it possible to change that, as long as the default behaviour doesn't change

Spikes042:
No problem @sgiehl, I can understand you have other things to do.

Yes, I thought of that and it shouldn't be a problem, as generally, a user should use `ParserAbstract::setVersionTruncation($type)` to set version truncation, which applies it globally.
In fact, as far as the user doesn't redefine `ParserAbstract::$maxMinorParts` in the child class, it will use the global (parent class) value

Spikes042:
Hello @sgiehl 
Is there a chance of this getting merged?

sgiehl (approved):

sgiehl:
Sorry for the delay and thanks again for the improvement.

Spikes042:
No problem at all. Really do appreciate

PR #6078, opened 2019-10-28
add new rules and tests for Nokia/Microsoft devices

mimmi20:


regexes/device/mobiles.yml
```diff
@@ -182,12 +194,121 @@ Nokia:
     - regex: 'TA-10(04|12|52)'
       model: '8'
     # 2018 Models
-    - regex: 'Nokia ([A-Za-z0-9\. ]+) Build'
-      model: '$1'
+    - regex: '(?:; )?Nokia 8\.1'
```
sgiehl: All `(?:; )?` in the beginning of the following regex are useless, as they are optional

mimmi20: removed

PR #6093, opened 2019-11-20
Adds detection for Servo browser engine

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -366,8 +366,6 @@
 - regex: '(?:Mobile|Tablet).*Firefox(?:/(\d+[\.\d]+))?'
   name: 'Firefox Mobile'
   version: '$1'
-  engine:
```
mimmi20: Could you add a new rule for the Firefox with Servo engine?

Tests/fixtures/smartphone-3.yml
```diff
@@ -4646,8 +4646,8 @@
     name: Firefox
     short_name: FF
     version: "39.0.0.0"
-    engine: Gecko
-    engine_version: ""
+    engine: WebKit
```
mimmi20: This is wrong. Firefox never uses the Webkit engine.

Tests/fixtures/desktop.yml
```diff
@@ -148,8 +148,8 @@
     name: Firefox
     short_name: FF
     version: "23.0"
-    engine: Gecko
-    engine_version: ""
+    engine: WebKit
```
mimmi20: This is wrong. Firefox never uses the Webkit engine.

mimmi20 (changes requested):
Please add a new rule for Firefox with Servo engine to make sure to not change the current detections.

mimmi20 (approved):

PR #6095, opened 2019-11-22

liviuconcioiu:


regexes/bots.yml
```diff
@@ -592,7 +592,15 @@
   producer:
     name: 'Google Inc.'
     url: 'http://www.google.com'
-    
+
+- regex: 'AppEngine-Google'
```
sgiehl: I'm not sure if we should count that as bot. Actually I think those useragents are of specific apps (identified by the app id)

regexes/bots.yml
```diff
@@ -625,7 +625,7 @@
     name: 'Visual Meta'
     url: 'https://www.shopalike.cz/'
 
-- regex: 'Googlebot(-Mobile|-Image|-Video|-News)?|Feedfetcher-Google|Google-Test|Google-Site-Verification|Google Web Preview|AdsBot-Google(-Mobile)?|Google-Adwords-Instant|APIs-Google|Mediapartners-Google|Google.*/\+/web/snippet|GoogleProducer|Google[ -]Publisher[ -]Plugin|Google-Shopping-Quality|Google-Adwords-DisplayAds|Google-Assess|Google-AdWords-Express|Google-speakr|Google-Read-Aloud'
+- regex: 'Googlebot(-Mobile|-Image|-Video|-News)?|Feedfetcher-Google|Google Web Preview|AdsBot-Google(-Mobile)?|APIs-Google|Mediapartners-Google|Google.*/\+/web/snippet|GoogleChirp|GoogleProducer|Google[ -]Publisher[ -]Plugin|Google-(AMPHTML|Test|Site-Verification|Shopping-Quality|Adwords-DisplayAds|Assess|AdWords-Express|Adwords-Instant|speakr|Read-Aloud|Youtube-Links|HotelAdsVerifier)?'
```
sgiehl: We should not add GoogleChirp here. It's not a bot, but a smart speaker directly streaming a PodCast. So actually it would be better to detect it as Media Player client.

See https://developers.google.com/search/docs/guides/podcast-management?hl=en

sgiehl (changes requested):

PR #6102, opened 2019-12-03
Add detection for Ordissimo OS, Ordissimo brand, Ordissimo browser

liviuconcioiu:


regexes/oss.yml
```diff
@@ -147,6 +147,10 @@
 ##########
 # Linux
 ##########
+- regex: '(Ordissimo|webissimo3)(?:/(\d+[\.\d]+))?'
+  name: 'Ordissimo'
+  version: '$2'
```
sgiehl: Are you sure the os and browser version are always the same? I would expect that to only be the browser revision. The os might not include a revision number to detect.

liviuconcioiu: From what I have seen in the user agent and what I've read on their website yes. I still don't know how the user agent looks like for smartphones for example.

sgiehl: for smarthpones they seem to use android and chrome. See https://www.handsetdetection.com/device-detection-database/devices/ordissimo/le-numero-1/

liviuconcioiu:
```
Mozilla/5.0 (X11; Linux i686 on x86_64; rv:52.0) Gecko/20100101 Firefox/52.0 Ordissimo/3.8.20+svn37598
Mozilla/5.0 (X11; Linux i686; rv:45.0) Gecko/20100101 Firefox/45.0 Ordissimo/3.8.6.6 svn37147
```
From what I see here, I think the best option will be to detect OS version 3.8.6.6 and 3.8.20 and browser version based on Firefox version. As you can see the newer OS version has an older Firefox version.

sgiehl: the firefox version is a newer one, as `3.8.20` is greater than `3.8.6.6`.
Maybe we should detect the version as the browser version, and don't detect a version for the OS?

sgiehl (approved):

PR #6106, opened 2019-12-09
Adds detection for DU Browser browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,11 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# DU Browser
+- regex: 'bdbrowser(?:_i18n)?(?:/(\d+[\.\d]+))?'
+  name: 'DU Browser'
```
sgiehl: Where do you have the information from that `bdbrowser` indicated it's `DU Browser`?

liviuconcioiu: https://github.com/Fyrd/caniuse/issues/4463

sgiehl: sorry for the long delay. Actually I wonder if we should group that with `Baidu Browser` as `DU Browser` seems to be their new name 🤔

PR #6121, opened 2019-12-11
Fix Sleipnir browser detection

liviuconcioiu:


sgiehl:
could you add a test here as well?

PR #6122, opened 2019-12-11
Fix NTENT Browser browser detection

liviuconcioiu:


sgiehl:
Same here. If it needs to be placed before Chrome, please add tests for that. The current test does not indicated that's needed...

liviuconcioiu:
Is not detected, as you can see here
`https://devicedetector.lw1.at/Mozilla%2F5.0%20(Android%205.1.1;%20U5151%20Build%2FLMY47V)%20AppleWebKit%2F537.36%20(KHTML,%20like%20Gecko)%20Chrome%2F54.0.2840.9%20NTENTBrowser%2F4.2.0.181%20(CellularOne-US)%20Mobile%20Safari%2F537.36`

PR #6126, opened 2019-12-11
Fix Kazehakase browser detection

liviuconcioiu:


sgiehl:
Kazehakase is a very old browser, that is imho not based on chrome, nor does it include `Chrome` in the useragent. If you have examples for that, please add them as tests...

liviuconcioiu:
My mistake with Chrome, just copy pasted from NTENT. I meant before Firefox.

sgiehl (approved):

PR #6131, opened 2019-12-11
Adds detection for Opera GX browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,19 @@
...
 ###############
 
+# Opera GX (https://www.opera.com/gx#start | http://ftp.opera.com/pub/opera_gx/)
+- regex: 'OPRGX(?:/(\d+[\.\d]+))?'
+  name: 'Opera GX'
+  version: '$1'
+  engine:
+    default: 'Blink'
+
+- regex: '(?:OPR/(60.0.3255.(50747|50962|51199)|62.0.3331.(105|117|132|52|81|96)|63.0.3368.(54789|54979|55362|55666|56078|56786|57388|57756|58152|66|71)|64.0.3417.(100|121|139|146)))'
```
sgiehl: Would you mind mentioning where you found the information that those specific versions are Opera GX only and can't be a normal Opera browser?

liviuconcioiu: [http://ftp.opera.com/pub/opera_gx/](http://ftp.opera.com/pub/opera_gx/) I have added the link to regexes/client/browsers.yml file.

sgiehl: I see the list of versions there. But I wonder if a normal opera might not have the same version number?

liviuconcioiu: From what I see here [http://ftp.opera.com/pub/opera/desktop/](http://ftp.opera.com/pub/opera/desktop/), it seems that they use the same version numbers for Opera too :(

sgiehl: Then lets remove the version specific detection of Opera GX

sgiehl (approved):

PR #6133, opened 2019-12-11
Adds detection for UC Browser Turbo browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -483,6 +483,14 @@
   name: 'UC Browser Mini'
   version: ''
 
+# UC Browser Turbo
+- regex: 'UCTurbo(?:[ /]?(\d+[\.\d]+))?'
+  name: 'UC Browser Turbo'
+  version: '$1'
+- regex: 'UC[ ]?Browser.* \(UCTurbo\)'
```
sgiehl: Could you add an test for that version?

sgiehl:
@liviuconcioiu  the test you added actually does match the regex `UC[ ]?Browser.* \(UCTurbo\)`. Do you have an example useragent for a test of that one as well?

PR #6244, opened 2020-04-13
Adds detection for Light browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,11 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# Light (https://sourceforge.net/projects/lightfirefox/)
+- regex: 'Light/(?:(\d+[\.\d]+))?'
```
sgiehl, line 9: If we require the version to be present we can simplify that to
```suggestion
- regex: 'Light/(\d+[\.\d]+)'
```

PR #6249, opened 2020-04-13
Adds detection for Origin browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,11 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# Origin (https://www.origin.com/)
```
sgiehl, line 8: what kind of browser is that? :thinking:

Findus23: I guess similar to steam the web browser running in their client displaying the shop pages (and linked sites)

PR #6250, opened 2020-04-13
Adds detection for PlayStation 4 Pro console

liviuconcioiu:


liviuconcioiu:
`  os:
    name: PlayStation
    short_name: PS3
    version: "4"
    platform: ""`
Here should be `4 Pro` but I don't know why it shows that.

sgiehl:
@liviuconcioiu  that actually the operating system. Guess that doesn't differ between 4 and 4 Pro, does it?

liviuconcioiu:
I've looked into the previous added consoles and I've seen their model added there. I just assumed that field should be `4 Pro`. On the other hand, PS4 runs on `Orbis OS`, and PS3 runs on `Cell OS`. Shouldn't those values be added as default, instead of the ones that are now?

sgiehl:
@liviuconcioiu feel free to change the OS names detected for PlayStation if they are wrong...

PR #6252, opened 2020-04-14
Adds detection for Steam browser

### LINKED ISSUE #6251: Add Steam Overlay
(Just a note as I don't have time to add it right now)

During Steam Games one can open a complete web browser in the game overlay. The user agent is
```
Mozilla/5.0 (Linux; U; X11; en-US; Valve Steam GameOverlay/1586022601; ) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.117 Safari/537.36

### PR DESCRIPTION
**Pull request by liviuconcioiu**
#6251

### CODE REVIEW

File: Parser/Client/Browser.php [hunk 1]
```diff
@@ -235,6 +235,7 @@ class Browser extends ClientParserAbstract
         'SU' => 'Super Fast Browser',
         'S3' => 'surf',
         'S0' => 'START Internet Browser',
+        'S4' => 'Steam',
```
sgiehl, line 238: Wondering if we maybe should name it `Steam Overlay` or `Steam GameOverlay` instead

liviuconcioiu: Long time since I played video games. Also, we could name it `Steam In-Game Overlay`. If so, I should also change it for `Origin`.

sgiehl: `Steam In-Game Overlay` sounds fine

PR #6266, opened 2020-05-01
Improves detection for Apple iPhone devices

liviuconcioiu:


regexes/device/mobiles.yml
```diff
@@ -467,72 +467,75 @@ Apple:
     - regex: '(?:Apple-)?iPh(?:one)?1[C,_]2'
       model: 'iPhone 3G'
       device: 'smartphone'
-    - regex: '(?:Apple-)?iPh(?:one)?2[C,_]1'
+    - regex: '(?:Apple-)?iPh(?:one)?2[C,_]1|3GS'
```
sgiehl, line 470: wondering if we maybe should have all detection look like this `; 3GS\)$` to ensure they won't match anything else. 🤔

liviuconcioiu: Is there anything wrong with the general regex, in other files? It seems to ignore the `;`.

sgiehl: maybe you might need to put some brackets around that. e.g. `|(?:; 3GS\)$)` 🤔

liviuconcioiu: `;? XR\)$` seems to work.

sgiehl: that's actually the same as leaving it away. Thinking a bit more about that, I guess the reason is here:
https://github.com/matomo-org/device-detector/blob/e8bee8ac311995daa078670e487239e3216cc9a7/DeviceDetector.php#L779-L788

So we actually only match if directly before the regex is no other letter, dash or underscore. That's to prevent we match something that is in between some other word or so

liviuconcioiu: ` 3GS\)$` would give a match for `; 3GS)`. Everything should be fine now.

PR #6288, opened 2020-05-05
Improves detection for Baidu Browser

liviuconcioiu:


Tests/fixtures/smartphone-11.yml
```diff
@@ -3674,17 +3674,17 @@
...
   client:
     type: browser
-    name: Chrome Webview
-    short_name: CV
-    version: "30.0.0.0"
-    engine: Blink
-    engine_version: ""
+    name: Baidu Browser
+    short_name: BD
+    version: ""
```
sgiehl, line 3679: Maybe we could detect the version here as well, if we change the regex to also handle `bdbrowser_i18n`

sgiehl (approved):

PR #6292, opened 2020-05-05
 Adds detection for multiple notebooks

liviuconcioiu:


misc/statistics.php
```diff
@@ -109,5 +110,6 @@ function format($str, $length)
     format($deviceTypes[8], $parsedUAs), getPercentage($deviceTypes[8], $parsedUAs),
     format($deviceTypes[9], $parsedUAs), getPercentage($deviceTypes[9], $parsedUAs),
     format($deviceTypes[10], $parsedUAs), getPercentage($deviceTypes[10], $parsedUAs),
+    format($deviceTypes[10], $parsedUAs), getPercentage($deviceTypes[13], $parsedUAs),
```
mimmi20, line 113: Should this not be `format($deviceTypes[13], $parsedUAs), getPercentage($deviceTypes[13], $parsedUAs)`

mimmi20 (changes requested):

sgiehl:
@liviuconcioiu  I'm not sure if we should add a new device type notebook 🤔  Nowadays most people are using notebooks, but most of the can't be detected as such and will be shown as desktop. Starting to detect some of them as notebooks now (and only those using facebook), might give a wrong insight

liviuconcioiu:
I see. I will move them to desktop then.

sgiehl:
@liviuconcioiu I'll merge this to a local branch an recreate the PR with the required changes...

PR #6310, opened 2020-06-16
Added a few Android devices

GregOriol:


Tests/fixtures/smartphone-4.yml
```diff
@@ -3002,11 +3002,31 @@
   os_family: Android
   browser_family: Android Browser
 - 
-  user_agent: Mozilla/5.0 (Linux; U; Android 4.4.2; en-tw; HTC One 801e Build/KOT49H) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Mobile Safari/534.30
```
sgiehl, line 3005: Any reason for removing the test case for `HTC One 801e`?

sgiehl (changes requested):

Tests/fixtures/smartphone-4.yml
GregOriol: Oups, didn't mean to remove it! Readded it, and enhanced the detection too.

sgiehl (approved):

PR #6327, opened 2020-07-17
Detect new browser: Blue Browser

sanchezzzhak:


sanchezzzhak:
Should we separate "Surf" for Android into a separate browser?

For example
Surf Mobile Browser

sgiehl:
@sanchezzzhak  I guess this mobile browser is not the mobile version from "surf" (https://surf.suckless.org/). If so, guess we should separate it...

sanchezzzhak:
I suggest excluding this browser indefinitely, as soon as more evidence is available, we will add it.

PR #6338, opened 2020-07-29
Adds detection for client DingTalk

coooold:
UserAgent: Mozilla/5.0 (iPad; CPU OS 11_2_6 like Mac OS X) AppleWebKit/604.5.6 (KHTML, like Gecko) Mobile/15D100 AliApp(**DingTalk/5.1.12**) com.laiwang.DingTalk/13441885 Channel/201200 Pad/iPad language/zh-Hans-CN UT4Aplus/0.0.6 WK


regexes/client/mobile_apps.yml
```diff
@@ -57,6 +57,11 @@
   name: 'WeChat'
   version: '$1'
 
+# DingTalk
+- regex: 'DingTalk/([^\)]+)'
```
sgiehl, line 61: Might be better to match the version number with digits instead of all chars except of `)`. Otherwise it might match too much if the identifier changes and is no longer surrounded by brackets...

```suggestion
- regex: 'DingTalk/([0-9\.]+)'
```

sgiehl (changes requested):

sgiehl (approved):

PR #6346, opened 2020-08-11
Detect notebook devices in Facebook useragents

sgiehl:
replaces/based on #6292


misc/statistics.php
```diff
@@ -109,5 +111,7 @@ function format($str, $length)
     format($deviceTypes[8], $parsedUAs), getPercentage($deviceTypes[8], $parsedUAs),
```
sanchezzzhak, line 111: let's improve here, we can do it universally

```php
<?php
/**
 * Device Detector - The Universal Device Detection library for parsing User Agents
 *
 * @link https://matomo.org
 * @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 */

require_once(__DIR__ . '/../vendor/autoload.php');

if (php_sapi_name() !== 'cli') {
    die("web not supported");
}
if (count($argv) != 2) {
    die("invalid arguments. Useage: php statistics.php filetoparse.txt");
}

use \DeviceDetector\Parser\Device\DeviceParserAbstract;

$deviceAvailableDeviceTypes = array_flip(DeviceParserAbstract::getAvailableDeviceTypes());

$parsedUAs = $unknownDeviceTypes =
$detectedBots = 0;

DeviceParserAbstract::setVersionTruncation(DeviceParserAbstract::VERSION_TRUNCATION_NONE);

$deviceTypes = (array_fill(0, count($deviceAvailableDeviceTypes), 0));

$startTime = microtime(true);

$handle = @fopen($argv[1], "r");

$parser = new DeviceDetector();

if ($handle) {
    while (($line = fgets($handle, 4096)) !== false) {

        if (empty($line)) {
            continue;
        }

        if ($parsedUAs > 0 && $parsedUAs % 80 == 0) {
            echo " $parsedUAs\n";
        }

        $parser->setUserAgent(trim($line));
        $parser->parse();

        echo '.';

        $parsedUAs++;

        if ($parser->isBot()) {
            $detectedBots++;
            continue;
        }

        if ($parser->getDevice() !== null) {
            $deviceTypes[$parser->getDevice()]++;
        } else {
            $unknownDeviceTypes++;
        }

    }
    if (!feof($handle)) {
        echo "Error: unexpected fgets() fail\n";
    }
    fclose($handle);
}

$timeElapsed = microtime(true) - $startTime;

function getPercentage($cur, $max)
{
    return format(round($cur * 100 / $max), '   ');
}

function format($str, $length)
{
    return sprintf("%" . strlen($length) . "d", $str);
}

$line = "----------------------------------------------\n";
$mask = "%24s %10s (%s%s) \n";
$reportStat = [];
$reportStat[] = sprintf($mask, 'Type', 'Count', 'Percent', '');
$reportStat[] = $line;
foreach ($deviceTypes as $deviceTypeId => $deviceCount) {
    $reportStat[] = sprintf(
        $mask,
        sprintf('%s:',mb_convert_case($deviceAvailableDeviceTypes[$deviceTypeId], MB_CASE_TITLE)),
        format($deviceCount, $parsedUAs),
        trim(getPercentage($deviceCount, $parsedUAs)),
        '%'
    );
}

$reportStat[] = sprintf(
    $mask,
    'Unknown',
    format($unknownDeviceTypes, $parsedUAs),
    trim(getPercentage($unknownDeviceTypes, $parsedUAs)),
    '%'
);
$reportStat[] = $line;

echo sprintf("

Parsed user agents:            %u

Total time elapsed:            %s
Average time per user agent:   %s

Detected Bots:    %s    (%s%%)

Detected device types:
%s
", $parsedUAs,
    round($timeElapsed, 2),
    round($timeElapsed / $parsedUAs, 6),
    format($detectedBots, $parsedUAs),
    getPercentage($detectedBots, $parsedUAs),
    implode('', $reportStat)
);

```
result 
image

regexes/device/notebooks.yml
```diff
@@ -0,0 +1,114 @@
...
+    - regex: 'FBMD/Aspire E5-421G;'
+      model: 'Aspire E5-421G'
+    - regex: 'FBMD/Z5WAL;'
+      model: 'Aspire E5-511'
+    - regex: 'FBMD/One S1003;'
+      model: 'One 10'
+
+Asus:
+  regex: 'FBMD/(?:K50IN|K54L|T100HAN|T103HAF|UX360CAK|X550LB|X553MA|X555LN|X556UQK);'
+  device: 'desktop'
```
sanchezzzhak, line 21: type notebook?

misc/statistics.php
sgiehl: could you do that in a new PR once that one was merged?

regexes/device/notebooks.yml
sgiehl: It used to be `notebook` in the original PR, but actually I think that wouldn't be good. We aren't able to say if it's a desktop or notebook in most cases. Having a new notebook type would look like we are able to detect that properly, but actually that only works for few cases...

Tests/fixtures/smartphone-1.yml
```diff
@@ -7288,6 +7268,26 @@
...
+  user_agent: Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/603.1.30 (KHTML, like Gecko) Mobile/14E5260b Version/10.3 Safari/8536.25 Mobicip/1947694464
+  os:
+    name: iOS
+    short_name: IOS
+    version: "10.3"
+    platform: ""
+  client:
+    type: browser
+    name: Mobicip
+    short_name: MO
```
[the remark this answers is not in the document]
sgiehl: could you do that in a new PR? `Mobicip` wasn't added in this PR, so be good not to change that here

PR #6369, opened 2020-09-18
Detect new brands: Ovvi, iBrit, Alcor, Fero and Detect devices for existing brands: Oppo, AllCall, Luna, Itbis, Ziox, Bq, Ginzzu, iPro, Lenovo, Odys, Highscreen

sanchezzzhak:
brand `Alldocube` merge by a `Cube`, since it is the same thing. https://github.com/matomo-org/device-detector/pull/6369/commits/14e6ae3ed5d5196fd1252a1a4debc46f847647d1


Parser/Device/DeviceParserAbstract.php
```diff
@@ -82,8 +82,8 @@ abstract class DeviceParserAbstract extends ParserAbstract
         'AK' => 'Akai',
         '1A' => 'Alba',
         'AL' => 'Alcatel',
+        '20' => 'Alcor',
         '4A' => 'Aligator',
-        '3A' => 'AllDocube',
```
sgiehl, line 86: Isn't AllDocube a separate brand anymore? 🤔

sanchezzzhak: Alldocube is the full name of the cube brand.

They were not divided into two companies.

I didn't find any information about splitting into two brands on the Internet

sanchezzzhak: Returned Alldocube back

sgiehl: I actually didn't do a longer research for `Cube` vs `Alldocube` but if it's actually the same brand, guess it would also be fine to merge them. But we maybe shouldn't remove the brand from the list. If anyone has stored the short codes, like Matomo does, it might fail fetching the brand name for older data. So we should leave it in there (with a comment) and use it no longer...

sanchezzzhak: I think we need to additionally check the device for images

iWork 8 Ultimate
image

iWork 10 Pro
image

Basically all new devices come with the Alldocube logo

sgiehl (approved):

Parser/Device/DeviceParserAbstract.php
sgiehl: Ok. For now let's just leave them as separate brands. We can change that later as well if we want to...

PR #6391, opened 2020-10-18
Detect new brands:  8848, Transpeed and Detect devices for existing brands: ZTE, Realme, Oppo, Vivo, Meitu, Chuwi, Motorola, Asus, Sony, Huawei, Xiaomi

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -7063,11 +7119,13 @@ Kaan:
 
 # Kazam
 Kazam:
-  regex: 'Kazam ([^;/)]+)( Build|\))|Trooper_X[0-9][0-9](?:[);/ ]|$)|Tornado 348'
+  regex: 'Kazam(?:[);/ ]|$)|Trooper_X[0-9][0-9](?:[);/ ]|$)|Tornado 348'
```
sgiehl, line 7122: Does that change make sense here? Don't think `Kazam` might be the end of the user agent. Maybe that should work:

```suggestion
  regex: 'Kazam(?:[^);/]+)(?:[);/ ]|$)|Trooper_X[0-9][0-9](?:[);/ ]|$)|Tornado 348'
```

Guess it should be change in the model regex as well

sanchezzzhak: I suggest replacing this rule with `Kazam(?:[^);/]+)(?:[);/ ]|$)|` the following  `Kazam[; _]|`  or `Kazam|`

sgiehl: that's fine. you can also optimize the model regex a few lines below

sanchezzzhak: @sgiehl done

sgiehl (approved):

PR #6399, opened 2020-10-27
Deprecate usage of short names

sgiehl:
In order to remove the short codes completely in maybe the next major release, we should start deprecating them first.
This PR will already remove the short_name usage from all tests. Additionally they won't be returned by `getInfoFromUserAgent` any longer.

### DIFF
--- a/misc/fileTest.php
+++ b/misc/fileTest.php
@@ -25,9 +25,7 @@
  * `php file-test.php /tmp/source-useragent.txt "not" "useragent" > /tmp/useragent-not-detected.txt`
  */
 use DeviceDetector\DeviceDetector;
-use DeviceDetector\Parser\Client\Browser;
 use DeviceDetector\Parser\Device\AbstractDeviceParser;
-use DeviceDetector\Parser\OperatingSystem;
 
 require __DIR__ . '/../vendor/autoload.php';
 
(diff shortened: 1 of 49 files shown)


misc/fileTest.php
```diff
@@ -25,9 +25,7 @@
  * `php file-test.php /tmp/source-useragent.txt "not" "useragent" > /tmp/useragent-not-detected.txt`
  */
 use DeviceDetector\DeviceDetector;
-use DeviceDetector\Parser\Client\Browser;
 use DeviceDetector\Parser\Device\AbstractDeviceParser;
-use DeviceDetector\Parser\OperatingSystem;
 
 require __DIR__ . '/../vendor/autoload.php';
```
sanchezzzhak, line 30: I'm not sure that ' use ' will work if autoload is enabled just below.

sanchezzzhak (changes requested):
bot tests should be fixed

misc/fileTest.php
sanchezzzhak: all is well, spl_autoload is started when the class is called, use ignores

sanchezzzhak (approved):

misc/fileTest.php
sgiehl: That works fine, as it will try to really use it the first time the class is used in the code...
It even doesn't fail if you define a class in a use statement that doesn't exist as long as you don't use it anywhere 🙈

PR #6405, opened 2020-10-27
Detect new brands: Artel, A1, Swisstone and Detect devices for existing brands: RCA Tables, ZTE, Digma, Wiko, Vivo, Google, Tone, CUBOT, OPPO, Aligator, Blu, Logicom, Huawei

sanchezzzhak:


sanchezzzhak:
@sgiehl Which size in rows is easier to accept?
1000 or 2500 lines?

sgiehl:
the smaller it is the faster I am with reviewing it...

regexes/device/mobiles.yml
```diff
@@ -14289,6 +14329,8 @@ Xiaomi:
       model: 'Redmi 10X Pro'
     - regex: 'Redmi K20 Pro(?:[);/ ]|$)'
       model: 'Redmi K20 Pro'
+    - regex: 'Redmi K30 5G(?:[);/ ]|$)'
+      model: 'Redmi K30 5G'
 
     - regex: 'MI PAD 4 PLUS(?:[);/ ]|$)'
       model: 'Mi Pad 4 Plus'
```
sanchezzzhak, line 14336: should we make the same format for MI * model names for the Xiaomi brand?
`Mi Pad 4 Plus`  => `MI Pad 4 Plus`

sgiehl: Sure. Sounds good

regexes/device/mobiles.yml
```diff
@@ -1090,237 +1090,241 @@ Arian Space:
...
   device: 'smartphone'
   models:
-    - regex: '(?:5022[EDX]|5070D)(?:[);/ ]|$)'
+    - regex: '5022[EDX]|5070D'
```
sgiehl, line 1096: The changes should actually be ok. But we need to keep in mind, that removing those checks means that it might create false positive matches for strings that start with the regexes but are longer. In this case for example `5070DF443`

sgiehl (approved):

PR #6406, opened 2020-10-27
Avoid using short names internally

sgiehl:


Parser/Device/AbstractDeviceParser.php
```diff
@@ -903,6 +903,24 @@ public static function getFullName(string $brandId): string
...
+     *
+     * @param string $brand brand name
+     *
+     * @return string
+     *
+     * @deprecated since 4.0 - short codes might be removed in next major release
+     */
+    public static function getShortCode(string $brand): string
+    {
+        if (\in_array($brand, self::$deviceBrands)) {
```
sanchezzzhak, line 917: ver1
```
$shortcode = \array_search($shortname2, $deviceBrands);
if(false !== $shortcode) {
  return shortcode;
}
return '';
```
ver2
Would I prefer to change the keys and values and use the $deviceBrands[$brandName] ?? ";
hash search is faster than array_search

sanchezzzhak (changes requested):

Parser/Device/AbstractDeviceParser.php
sgiehl: Ok. So to make that even shorter we could have:
```
return \array_search($brand, $deviceBrands) ?: ''
```

or:

```
$flip = \array_flip(self::$deviceBrands);
return $flip[$brand] ?? '';
```

What would you prefer?

sanchezzzhak: it is best to set the array in the properties as is
https://github.com/matomo-org/device-detector/blob/master/Parser/Device/AbstractDeviceParser.php#L79-L90

...
'Apple' => 'AP',
'Samsung' => 'SU',
'Huawei' => 'HW',
...

and use `return $flip[$brand] ?? '';```

If that doesn't work then first variant

sgiehl: Wouldn't personally flip the definition array, as at some point we might remove the short names. And it's easier to simply remove the keys then instead of another flip

sanchezzzhak (dismissed):

sanchezzzhak (approved):

PR #6409, opened 2020-11-02
Update caching examples and composer add require example

manavo:
The Doctrine cache requires the `DoctrineBridge` bridge to actually work


sgiehl (approved):

sgiehl:
thanks @manavo for updating that

manavo:
Thanks for the quick merge!

PR #6426, opened 2020-11-13
 Added includes inside yml & split models into files 

sanchezzzhak:
…brands Arcatel, Apple, Asus, BQ, Huawei, Samsung, LG, Sony-*, vivo, Xiaomi, ZTE


sanchezzzhak:
@sgiehl I suggest we discuss whether we need this functionality?

sgiehl:
Not sure if we should do that. In general it might be a bit more cleaner to split the big ones into separate files. On the other side there are various ports to other languages that depend on the regex files, so splitting them would mean breaking those ports. Also that means reading more files when parsing a useragent.

PR #6434, opened 2020-11-20
Introduce Yaml Lint check to prevent errors in yml files

### LINKED ISSUE #6433: Add a test using the Symfony yaml parser to parse all files
to avoid typos like #6432 in the future

sgiehl:
fixes #6433


sgiehl:
@Findus23 I've added a Linter instead of adding a test with symfony. Tried the problem we had before and it's correctly marked as error. See https://github.com/matomo-org/device-detector/pull/6434/commits/780b749e245e5989f8d33b88cf0e2b87ede266f9
Guess that should be enough to prevent further problems like this, right?

Findus23 (dismissed):
This might even be better. Thanks for the change.

Findus23:
Not sure how helpful error messages like this are:

```text
======================
= Linting YAML files =
======================
Error: syntax error: expected <block end>, but found '<block sequence start>' (syntax)
Error: Process completed with exit code 1.
```

sanchezzzhak:
@sgiehl Is it possible to make a more detailed error reports in YAML Lint?

Findus23:
Interestingly running `yamllint regexes/**/*.yml` locally gives nice error messages.

sgiehl:
as @Findus23 mentioned. The errors should be highlighted in the diff on github. Guess there's not much that can be changed on the error messages. But imho that should be enough... we should at least now see if any yml file is broken

Findus23 (approved):
Sounds good to me

sanchezzzhak (approved):

PR #6460, opened 2020-11-30
Adds detection for Phoenix Browser

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -234,6 +234,7 @@ class Browser extends AbstractClientParser
         'PW' => 'Palm WebPro',
         'PA' => 'Palmscape',
         'PX' => 'Phoenix',
+        'PB' => 'Phoenix Browser',
```
sanchezzzhak, line 237: shortcode must be added to the chrome browser family
https://github.com/matomo-org/device-detector/blob/77af448d5a25edb602f1353c79d9b0bb57e3fa56/Parser/Client/Browser.php#L327-L337

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #6466, opened 2020-12-02
Run tests on PHP 8.0

### LINKED ISSUE #6465: PHP 8.0
Should tests also be run on PHP 8.0?

liviuconcioiu:
ref  #6465


sgiehl:
@liviuconcioiu seems we need to include the php 8 build like the nightly build for now, as the dependencies currently seem not to be marked as compatible with php 8

sgiehl:
We've switched to GitHub actions. Tests are now running on PHP 8 as well

PR #6472, opened 2020-12-03
Adds detection for Chim Lạc browser

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -90,6 +90,7 @@ class Browser extends AbstractClientParser
         'CN' => 'CoolNovo',
         'CO' => 'CometBird',
         'CB' => 'COS Browser',
+        'C3' => 'Chim Lạc',
```
sanchezzzhak, line 93: I don't think it's a good idea to use Vietnamese characters
```yml
Chim Lac
```
@sgiehl do you think we should replace "Chim Lạc" with "Chim Lac"?

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #6475, opened 2020-12-03
Adds detection for Bo Browser

liviuconcioiu:


sanchezzzhak:
BoBrowser, It is an undesirable software product, many sites say that it is a virus/adware
@sgiehl do we add dishonest browsers?

sgiehl:
Die a quick search as well, and seems @sanchezzzhak is right. Wouldn't added detection of software that is obviously bad

PR #6476, opened 2020-12-03
Adds detection for Beagle Browser

liviuconcioiu:


sanchezzzhak:
I think this browser is adware software. need more time to check

sanchezzzhak:
I think it is not necessary to add this browser.
many sites indicate that this is adware.

PR #6487, opened 2020-12-03
Adds detection for Freebox operating system

liviuconcioiu:


sanchezzzhak:
https://fr.wikipedia.org/wiki/Freebox
most likely, this is a device similar to mikrotik with Linux under the hood

sgiehl:
> https://fr.wikipedia.org/wiki/Freebox
> most likely, this is a device similar to mikrotik with Linux under the hood

@sanchezzzhak does that mean you wouldn't add it as detected operating system?

sanchezzzhak:
We can add this as an operating system, according to the information https://free.fr/assistance/5049.html

sgiehl (approved):

PR #6497, opened 2020-12-06
Call parse() whenever it's required

renanbr:
The intention is to avoid missing the `parser()` call.

I didn't found if this subject was already discussed in this project.

I'm aware that this may be a BC break for some, as an alternative solution, we could throw an exception (still a potential BC break), or trigger an error (less intrusive) if the user-agent was not yet parsed when calling these functions.

### DIFF
--- a/DeviceDetector.php
+++ b/DeviceDetector.php
@@ -213,12 +213,16 @@ class DeviceDetector
     {
         foreach (AbstractDeviceParser::getAvailableDeviceTypes() as $deviceName => $deviceType) {
             if (\strtolower($methodName) === 'is' . \strtolower(\str_replace(' ', '', $deviceName))) {
+                $this->parse();
+
                 return $this->getDevice() === $deviceType;
             }
         }
 
         foreach (self::$clientTypes as $client) {
             if (\strtolower($methodName) === 'is' . \strtolower(\str_replace(' ', '', $client))) {
+                $this->parse();
+
                 return $this->getClient('type') === $client;
             }
         }
@@ -326,6 +330,8 @@ class DeviceDetector
      */
     public function isBot(): bool
     {
+        $this->parse();
+
         return !empty($this->bot);
     }
 
@@ -350,6 +356,8 @@ class DeviceDetector
      */
     public function isMobile(): bool
     {
+        $this->parse();
+
         // Mobile device types
         if (\in_array($this->device, [
             AbstractDeviceParser::DEVICE_TYPE_FEATURE_PHONE,
@@ -397,6 +405,8 @@ class DeviceDetector
      */
     public function isDesktop(): bool
     {
+        $this->parse();
+
         $osName = $this->getOsAttribute('name');
 
         if (empty($osName) || self::UNKNOWN === $osName) {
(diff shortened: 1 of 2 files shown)


renanbr:
These two tests fail because they test exactly the opposite of the intention of this PR:

- DeviceDetector\Tests\DeviceDetectorTest::testGetOs
- DeviceDetector\Tests\DeviceDetectorTest::testGetClient

I can adapt it if the idea of the PR gets accepted.

DeviceDetector.php
```diff
@@ -213,12 +213,16 @@ public function __call(string $methodName, array $arguments): bool
     {
         foreach (AbstractDeviceParser::getAvailableDeviceTypes() as $deviceName => $deviceType) {
             if (\strtolower($methodName) === 'is' . \strtolower(\str_replace(' ', '', $deviceName))) {
+                $this->parse();
```
sgiehl, line 216: that would actually not be needed, as `getDevice` would already trigger a `parser`

DeviceDetector.php
```diff
@@ -213,12 +213,16 @@ public function __call(string $methodName, array $arguments): bool
...
+
                 return $this->getDevice() === $deviceType;
             }
         }
 
         foreach (self::$clientTypes as $client) {
             if (\strtolower($methodName) === 'is' . \strtolower(\str_replace(' ', '', $client))) {
+                $this->parse();
```
sgiehl, line 224: Same here. `parse` would be indirectly triggered by `getClient`

DeviceDetector.php
```diff
@@ -470,6 +486,8 @@ public function getDevice(): ?int
      */
     public function getDeviceName(): string
     {
+        $this->parse();
```
sgiehl, line 489: wouldn't be needed

sgiehl (changes requested):
@sanchezzzhak  what do you think in general about those changes?

DeviceDetector.php
sanchezzzhak: I don't think it's a good idea to run parser in every methods isBot isTablet etc.
we have a special method to check whether the parser method was started

https://github.com/matomo-org/device-detector/blob/master/DeviceDetector.php#L541-L555

I think it should just be a check, if the parser was not called, then a fatal exception
e.g.
```php
foreach (AbstractDeviceParser::getAvailableDeviceTypes() as $deviceName => $deviceType) {
            if (\strtolower($methodName) === 'is' . \strtolower(\str_replace(' ', '', $deviceName))) {
                if ( false === $this->isParse()) {
                      throw new \Exception("method parse() has not been called before");
                }
```
of course, if our goal is to force the user to call the parse() method

renanbr:
Hello, just giving you an update on this.

We solved the issue on our side by proxying `DeviceDetector`

Implementation details: https://github.com/acsiomatic/device-detector-bundle/pull/18

sgiehl:
@renanbr thanks for the update. That sounds like a better solution.
Even though some users might expect an automatic parse if methods that require a parse are called, but there might also be cases where that should not happen. Especially when you let your own classes inherit from DeviceDetector in order to replace some parts. Not doing the parse makes it more flexible. I'll close this PR then.

PR #6498, opened 2020-12-06
Adds detection for bots

liviuconcioiu:


regexes/bots.yml
```diff
@@ -2118,6 +2124,67 @@
...
+
+- regex: 'adbeat'
+  name: 'Adbeat'
+  category: 'Crawler'
+  url: 'https://www.adbeat.com/operation_policy'
+  producer:
+    name: 'PPC Labs LLC'
+    url: 'https://www.adbeat.com/'
+
+- regex: 'BW'
```
sanchezzzhak, line 2143: 2 characters is very small, you need to add one more condition to avoid collisions in the future.
e.g.
```yml
BM/(?:(\d+[\.\d]+))
```

sanchezzzhak (changes requested):

sgiehl:
@sanchezzzhak  guess that one looks good to merge now, right?

sanchezzzhak (approved):

sanchezzzhak:
> @sanchezzzhak guess that one looks good to merge now, right?

Yes, it can be merged into the master, I just checked

PR #6521, opened 2020-12-11
Improves detection for Samsung TVs

liviuconcioiu:


Tests/fixtures/tv.yml
```diff
@@ -6440,7 +6440,7 @@
     model: Z10
   os_family: Android
   browser_family: Chrome
--
+-
```
sanchezzzhak, line 6443: remove space char symbol

sanchezzzhak (approved):

PR #6546, opened 2020-12-20
Adds detection for tvOS and removes Apple TV

liviuconcioiu:


Parser/OperatingSystem.php
```diff
@@ -270,7 +269,7 @@ public static function getNameFromId(string $os, ?string $ver = null): ?string
      */
     protected function parsePlatform(): string
     {
-        if ($this->matchUserAgent('arm|aarch64')) {
+        if ($this->matchUserAgent('arm|aarch64') || $this->matchUserAgent('Apple TV|AppleTV')) {
```
sanchezzzhak, line 272:
```php
        if ($this->matchUserAgent('arm|aarch64|Apple ?TV')) {
```

sanchezzzhak (changes requested):

Parser/OperatingSystem.php
```diff
@@ -39,7 +39,6 @@ class OperatingSystem extends AbstractParser
         'AIX' => 'AIX',
         'AND' => 'Android',
         'AMG' => 'AmigaOS',
-        'ATV' => 'Apple TV',
```
sgiehl, line 42: We shouldn't remove anything from this list. If someone has stored that short code and tries to look up the name it will fail afterwards. Might be better to keep using the short code, but give it a new name.

regexes/oss.yml
```diff
@@ -300,6 +300,18 @@
...
+##########
+# tvOS (https://developer.apple.com/tvos/)
+##########
+- regex: 'Apple TV.*CPU OS ((?:9|10|11|12|13|14)_\d+(?:[_\.]\d+)*)'
+  name: 'tvOS'
+  version: '$1'
+
+- regex: 'iOS; ((?:9|10|11|12|13|14).\d+(?:[_\.]\d+)*) AppleTV'
+  name: 'tvOS'
+  version: '$1'
```
sgiehl, line 312: Isn't there still a AppleTV detection in that file? Guess that should be removed and covered with those detections

sgiehl (changes requested):

regexes/oss.yml
liviuconcioiu: It can't be removed because there are two different types of user agents, only if we can combine those two regexes.
`AppleCoreMedia/1.0.0.12B466 (Apple TV; U; CPU OS 8_1_3 like Mac OS X; en_us)
XBMC/PRE-11.0 Git:20110623-62171b3 (iOS; 11.0.0 AppleTV2,1; http://www.xbmc.org)`

Parser/OperatingSystem.php
liviuconcioiu: I've added back Apple TV, but it fails, because there are no test user agents for this OS anymore.

regexes/oss.yml
sgiehl: I was referring to this one:
https://github.com/matomo-org/device-detector/blob/cdf7d731e0d6b4ff2f8791f1debe6612649d4a0f/regexes/oss.yml#L685-L687

Parser/OperatingSystem.php
sgiehl: @liviuconcioiu are all Apple TVs running tvOS? If so, we should not add a new OS, but reuse the `ATV` and change the name to `tvOS`

liviuconcioiu: The 1st generation came with a OS based on Mac OS X 10.4 Tiger. 2nd, 3rd were based on iOS. After 4th there are based on tvOS.

Findus23 (commented):
Just to be safe it would be great if @brototyp could take a look at this change before it is merged to avoid issues with the iOS SDK.

sanchezzzhak (dismissed):

regexes/oss.yml
sanchezzzhak: here you can improve
```yml
- regex: 'iOS; ((?:9|1[0-4]).\d+(?:[_\.]\d+)*) AppleTV'
  name: 'tvOS'
  version: '$1'
  ```

sanchezzzhak:
Starting from the 9th version of AppleTV, the devices run under the tvOS operating system, and this PR takes this into account.
https://en.wikipedia.org/wiki/TvOS

sgiehl (dismissed):

sanchezzzhak:
@liviuconcioiu Hi, fix PR for the merger as soon as there time.
Thanks

sanchezzzhak (approved):

PR #6548, opened 2020-12-20
Adds detection for Odin browser

liviuconcioiu:


sanchezzzhak (approved):
Browser exclusively for Hisense TVs

PR #6590, opened 2021-01-20
Improves detection for Oregano browser

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -231,6 +231,7 @@ class Browser extends AbstractClientParser
         'OY' => 'Origyn Web Browser',
         'OV' => 'Openwave Mobile Browser',
         'OW' => 'OmniWeb',
+        'YX' => 'Onyx',
```
sanchezzzhak, line 234: https://en.wikipedia.org/wiki/Oregano_(web_browser)
it would probably be more correct to leave the name `Oregano Browser`

sanchezzzhak (changes requested):

Parser/Client/Browser.php
liviuconcioiu: Then, I think is much better to leave it as `Oregano`, and improve that detection. Because is the same browser.

sanchezzzhak: you are right, it is better to make one name `Oregano`

regexes/client/browsers.yml
```diff
@@ -5,6 +5,11 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# Oregano Browser (https://oregan.net/)
+- regex: 'OreganMediaBrowser(?:/(\d+[\.\d]+))?'
+  name: 'Oregano Browser'
+  version: '$1'
+
```
sanchezzzhak, line 12: if possible, move the rule closer to the oregano browser

sanchezzzhak (changes requested):

regexes/client/browsers.yml
liviuconcioiu: I've joined both regexes.

sanchezzzhak (approved):

PR #6596, opened 2021-01-22
Adds detection for Fenix operating system and Biyubi browser

liviuconcioiu:


Parser/OperatingSystem.php
```diff
@@ -151,7 +152,7 @@ class OperatingSystem extends AbstractParser
         'GNU/Linux'             => [
             'LIN', 'ARL', 'DEB', 'KNO', 'MIN', 'UBT', 'KBT', 'XBT', 'LBT', 'FED',
             'RHT', 'VLN', 'MDR', 'GNT', 'SAB', 'SLW', 'SSE', 'CES', 'BTR', 'SAF',
-            'ORD', 'TOS', 'RSO', 'DEE', 'FRE', 'MAG',
+            'ORD', 'TOS', 'RSO', 'DEE', 'FRE', 'MAG', 'FEN',
```
sanchezzzhak, line 155: The operating system is built from scratch, at least I did not find that it was created on the basis of Unix.
``` 'Real-time OS'          => ['MTK', 'TDX', 'MRE',  'FEN'],```

sanchezzzhak (changes requested):

Parser/OperatingSystem.php
Findus23: It is called https://fenixlinux.com and they say `Fenix OS es una distribución Linux` on their website, so I assume that they use a normal Linux kernel and are just a distribution like Debian.

sanchezzzhak (approved):

PR #6605, opened 2021-01-27
Adds detection for xStand browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,11 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# xStand (https://www.adnx.com/)
+- regex: 'xStand/(\d+[\.\d]+)'
+  name: 'xStand'
+  version: '$1'
```
sanchezzzhak, line 11: https://www.adnx.com/faqs
```
xStand is using Apple Webkit technology. On the Mac version, we are still using the classic webview but on the iPad version, we are already using the advanced WKwebview version.
```

default engine webkit

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #6613, opened 2021-01-29
Detect browsers: SP Browser, Opera Mini iOS and Detect apps: TikTok, RoboForm, Naver, Papers, Pinterest, Copied, Pic Collage, Slack

sanchezzzhak:


Tests/fixtures/mobile_apps.yml
```diff
@@ -719,3 +703,51 @@
...
+-
+  user_agent: Slack/396839 CFNetwork/889.9 Darwin/17.2.0
+  os:
+    name: iOS
+    version: "11.1"
+    platform: ""
+  client:
+    type: mobile app
+    name: Slack
+    version: ""
```
liviuconcioiu, line 747: Slack version is not detected.

regexes/client/mobile_apps.yml
```diff
@@ -76,7 +76,7 @@
   version: '$1'
 
 # Pinterest
-- regex: 'Pinterest(?:/([\d\.]+))?'
+- regex: 'Pinterest(?: for iOS)?(?:/([\d\.]+))?'
```
liviuconcioiu, line 79: Can you also add detection for `Pinterest for Android/9.0.4 (K83CA; 9)`

Tests/fixtures/mobile_apps.yml
sanchezzzhak: this is done intentionally

Slack/396839 the versions don't match the appstore

Tests/fixtures/mobile_apps.yml
```diff
@@ -623,3 +623,131 @@
...
+    type: ""
+    brand: Apple
+    model: ""
+  os_family: iOS
+  browser_family: Unknown
+-
+  user_agent: Pinterest for iOS/7.11 (iPhone7,2; 12.1.4)
+  os:
+    name: iOS
+    version: "7.11"
```
liviuconcioiu, line 726: iOS version should be `12.1.4`

regexes/oss.yml
```diff
@@ -325,6 +329,14 @@
 ##########
 # iOS
 ##########
+- regex: 'Pinterest for iOS/.+; (\d(?:[\d\.]+)?)\)$'
+  name: 'iOS'
+  version: '$1'
+
+- regex: 'CFNetwork/.+ Darwin/20.2.0(?!.*x86_64)'
```
liviuconcioiu, line 336: Those 3 lines are duplicates with the ones below them.

regexes/client/mobile_apps.yml
```diff
@@ -271,3 +271,42 @@
   regex: 'Procast/?([\d\.]+)?'
   name: 'Procast'
   version: '$1'
+-
+  regex: 'TikTok/(\d+[\.\d]+) CFNetwork'
+  name: 'TikTok'
+  version: ''
```
sgiehl, line 277: Isn't TikTok (and maybe others) available for Android as well? Those regexes would only match the iOS versions...

sanchezzzhak: I do not have a user agent for android

sgiehl: Don't have it either. Maybe we could simply remove the `CFNetwork` at least for those apps we now there is an Android version maybe?

sanchezzzhak: I removed the prefix `CFNetwork` for tiktok

sgiehl (approved):

PR #6629, opened 2021-02-06
Adds detection for Android Go operating system

liviuconcioiu:


regexes/oss.yml
```diff
@@ -115,6 +115,13 @@
   name: 'FydeOS'
   version: ''
 
+##########
+# Android Go (https://www.android.com/versions/go-edition/)
+##########
+- regex: '(?:(?:Orca-)?Android|Adr)[ /](?:[a-z]+ )?(\d+[\.\d]*).+(?:BLOCK 3 mini|Spark Plus 3)'
+  name: 'Android Go'
+  version: '$1'
+
```
sanchezzzhak, line 124: I don't think it's a good idea to define Android Go by the model name. 

There are many devices running this version of Android, we just don't physically list them all.

PR #6641, opened 2021-02-13
Adds detection for Qwant Mobile iOS browser

liviuconcioiu:


sanchezzzhak:
@sgiehl Is it acceptable to split the browser into two names? 
Similar precedent: `Opera Mini iOS`, `Firefox Mobile iOS`

sgiehl:
@sanchezzzhak  sorry for the delay. Good question. I guess I actually would say no. We should detect that as one, to keep it simple. There is actually no benefit in splitting it. We maybe should even consider removing the other iOS browser versions. Or do you have another opinion on that?

sanchezzzhak:
I am in favor of not specifying prefixes

regexes/client/browsers.yml
```diff
@@ -635,6 +635,11 @@
     default: 'Blink'
 
 # Qwant Mobile
+- regex: 'QwantiOS/(\d+[\.\d]+)'
+  name: 'Qwant Mobile iOS'
```
sanchezzzhak, line 639: change name: `Qwant Mobile`

sanchezzzhak (changes requested):

sanchezzzhak (approved):

Issue #6171, opened 2020-02-12
unknown browsers list


sanchezzzhak:
* [x] #6841

* header privatebrowser.securebrowser.com.klar
* [x] Mozilla/5.0 (Linux; Android 5.1) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Private Browser/4.0 Chrome/46.0.2490.76 Mobile Safari/537.36
* Mozilla/5.0 (Linux; Android 10) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Private Browser/4.0 Chrome/91.0.4472.120 Mobile Safari/537.36

* [x] Mozilla/5.0 (Linux; Android 5.1.1) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Arvin/2.0.0 Chrome/60.0.3112.107 Mobile Safari/537.36

**upd 11.11.22k**

Stargon
https://play.google.com/store/apps/details?id=net.onecook.browser

* [x] Mozilla/5.0 (Linux; Android 4.4.4; One Build/KTU84L.H4) AppleWebKit/537.36 (KHTML, like Gecko) Stargon/2.8.2 Chrome/33.0.0.0 Mobile Safari/537.36

https://slimerjs.org 
https://github.com/laurentj/slimerjs
* [x] Mozilla/5.0 (X11; Linux x86_64; rv:52.0) Gecko/20100101 SlimerJS/0.10.3

PR #6647, opened 2021-02-13
Adds detection for Arvin browser

liviuconcioiu:
#6171


sanchezzzhak:
@sgiehl Do we add this to the definition?

I tested this app and the video ads were in search before showing the result. (On the store's page, it is reported that there is advertising, everything is honest here)
The promised built-in proxy is present (here, too, they did not deceive)

sanchezzzhak (dismissed):

sgiehl (dismissed):

sgiehl (approved):

PR #6648, opened 2021-02-13
Adds detection for CGN mobile app

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,13 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# CGNBrowser (https://www.cgn.it/)
```
sanchezzzhak, line 8: it doesn't look like a browser for constant use
image
we can add this as an `mobile app`,  name:  `Platform CGN`

Parser/Client/Browser.php
```diff
@@ -80,6 +80,7 @@ class Browser extends AbstractClientParser
         'CA' => 'Camino',
         'CL' => 'CCleaner',
         'C0' => 'Centaury',
+        'CG' => 'CGN Browser',
```
sanchezzzhak, line 83: remove

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #6654, opened 2021-02-13
Adds detection for Thunder app

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,13 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# Thunder (https://xl11.xunlei.com/)
```
sanchezzzhak, line 8: I don't think this is a browser for surfing the internet
We can add this as an mobile-app in this PR

sanchezzzhak (approved):

PR #6658, opened 2021-02-13
Adds detection for Deledao browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -5,6 +5,13 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+# Deledao (https://www.deledao.com/)
+- regex: '(?:DeledaoPersonal|DeledaoFamily)/(\d+[\.\d]+)'
```
sanchezzzhak, line 9: I think this is not a browser but a browser extension
@sgiehl Do we need to classify this separately as a browser?

liviuconcioiu: [https://www.deledao.com/family-download](https://www.deledao.com/family-download)

sanchezzzhak: thank you for the link
> (Psst...it's an AI-powered browser!)
It’s built from the same code that Chrome is based on. Anything that Chrome is compatible with, Deledao Junior is too.

sanchezzzhak (approved):

PR #6665, opened 2021-02-14
Adds detection for JavaFX WebView

liviuconcioiu:


sgiehl:
@sanchezzzhak  Looks good to me. Should we add a `WebView` to the name, so it's clear it's an embedded browser? Don't have a strong opinion on that...

sanchezzzhak:
I looked at 3 projects, how they define this string.
51degrees: does not detect
WhichBrowser: JavaFX
udger: JavaFX

sanchezzzhak (dismissed):

sanchezzzhak (approved):

PR #6681, opened 2021-02-24
Improves detection for Maxthon browser

liviuconcioiu:


sanchezzzhak:
https://en.wikipedia.org/wiki/Maxthon#Version_5.x
> Maxthon 5 (also known as MX5) was released on October, 2016.[66] In the 5.2.x versions, the core of the browser is updated to the Chromium 61 branch point which suggests that the core is using Blink, a fork of WebKit.

all right

sanchezzzhak (approved):

PR #6686, opened 2021-02-24
Adds detection for Mandarin browser

liviuconcioiu:


sanchezzzhak:
this browser has a dubious reputation. 
By reviews:
1 steals personal data
2 loads CPU
3 mining coins
I would like to test this browser on a virtual machine.

@liviuconcioiu  don't update this PR, I'll let you know as soon as I check it.

sanchezzzhak (dismissed):
I checked this application with different scanners and found nothing.

sgiehl:
If this is a Chromium based browser we should add it to the Chrome browser family.

sanchezzzhak (approved):

PR #6690, opened 2021-02-25
Detect devices for brands: LG, Fujitsu, Chuwi, Sharp, FireFly Mobile, and create ExtraYaml check is not key number

sanchezzzhak:


sanchezzzhak:
@sgiehl need to escape only the float values in the key?

sgiehl:
> @sgiehl need to check only the float key?

Actually not sure, if that might also be the case for decimals. The error with symfony was:
```
PHP message: PHP Fatal error:  Uncaught Symfony\Component\Yaml\Exception\ParseException: Numeric keys are not supported. Quote your evaluable mapping keys instead at line 1305 (near "5.2: 'Blink'"). in /var/www/devicedetector/vendor/symfony/yaml/Parser.php:224
```

Tests/DeviceDetectorTest.php
```diff
@@ -460,6 +460,63 @@ public function testCheckRegexRestrictionEndCondition(): void
...
+            \glob(__DIR__ . '/../regexes/*.yml')
+        );
+        $yamlSymfony = new Symfony();
+
+        foreach ($files as $file) {
+            $yamlSymfony->parseFile($file);
+        }
+
+        $this->markTestIncomplete();
+    }
```
sanchezzzhak, line 518: @sgiehl the error occurs only on keys with fractions. on keys `5.2.0`, the parser does not crash.

which of these tests should we leave?

sgiehl: Guess would be enough to keep the one that directly uses the YAML parser from Symfony.

Tests/DeviceDetectorTest.php
```diff
@@ -460,6 +460,63 @@ public function testCheckRegexRestrictionEndCondition(): void
...
+
+    /**
+     * check the Symfony parser for fixtures parsing errors
+     */
+    public function testSymfonyParser(): void
+    {
+        $files       = \array_merge(
+            \glob(__DIR__ . '/../regexes/client/*.yml'),
+            \glob(__DIR__ . '/../regexes/client/*.yml'),
```
sgiehl, line 508:
```suggestion
            \glob(__DIR__ . '/../regexes/device/*.yml'),
```

Tests/DeviceDetectorTest.php
```diff
@@ -460,6 +460,25 @@ public function testCheckRegexRestrictionEndCondition(): void
...
+            \glob(__DIR__ . '/../regexes/device/*.yml'),
+            \glob(__DIR__ . '/../regexes/*.yml')
+        );
+        $yamlSymfony = new Symfony();
+
+        foreach ($files as $file) {
+            $yamlSymfony->parseFile($file);
+        }
+
+        $this->markTestIncomplete();
```
sgiehl, line 479: Guess it shouldn't be marked as incomplete, right?
If you want to avoid the test being marked as risky (due to no assertion) you can add a 
`$this->expectNotToPerformAssertions();`

sanchezzzhak: done

sgiehl (approved):

PR #6700, opened 2021-03-06
Detect new brands: Suzuki, Nabi, Gresso, G-Touch, FaRao Pro, StrawBerry, Smartab, Subor, SuperTab, Pendoo, Engel and Detect devices for existing brands: Sugar, TrekStor, Ghia, Lenovo, Freetel, Lark, Zidoo, Sony, Siragon, SPC, Karbonn, Bitel, Cloudfone, Navon, Ritmix, Starlight, Zuum, Blu, Dune HD, Philco

sanchezzzhak:
this PR merge after #6695


regexes/device/mobiles.yml
```diff
@@ -18784,14 +18848,17 @@ Maxtron:
...
   device: 'smartphone'
   models:
     - regex: '(PCS01|PH350B)(?:[);/ ]|$)'
       model: '$1'
     - regex: '(PTB7R)(?:[);/ ]|$)'
       device: 'tablet'
       model: '$1'
+    - regex: 'PHILCO Android TV(?:[);/ ]|$)'
+      device: 'tv'
+      model: 'General Android TV'
```
sgiehl, line 18861: Not sure if should name such devices at all. Leaving the name empty would report them as `Unknown`, which would actually be true, as we don't know the exact model 🤔

sanchezzzhak: I will correct for an empty value

sgiehl (approved):

PR #6723, opened 2021-04-19
 Detect devices for existing brands: Xiaomi, TCL, Alcatel, Hisense, Huawei, Digma, Irbis, Cricket, Hyundai, Apple, Lenovo,  AT&T, LG, Motorola, Kyocera, Lanix, OnePlus, iPro, Dell, Zuum, Polaroid, CHIA and Detect library Akka HTTP

sanchezzzhak:


regexes/client/browsers.yml
```diff
@@ -1166,6 +1166,11 @@
   version: '$1'
   engine:
     default: 'Blink'
+- regex: 'Opera%20Touch/(\d+[\.\d]+)? CFNetwork/.+Darwin/.+(?!.*x86_64)'
```
sgiehl, line 1169: Shouldn't this be enough? Or do you match the other parts for the Engine?

```suggestion
- regex: 'Opera%20Touch/(\d+[\.\d]+)?'
```

regexes/oss.yml
```diff
@@ -649,6 +657,10 @@
   name: 'iOS'
   version: ''
 
+- regex: 'iOS; (?!\d+)'
```
sgiehl, line 660: What is the `(?!\d+)` good for?

regexes/client/browsers.yml
sanchezzzhak: this rule is only for iOS mobile devices, where the browser engine is always webkit

sgiehl: ok 👍

regexes/oss.yml
sanchezzzhak: I'll redo everything.

sgiehl (approved):

PR #6755, opened 2021-05-27
Detect new brand: New Balance, Astro, Cobalt, FMT, mPhone, WIWA and Detect devices for existing brands: Asus, Maxwest, Intex, Celkon, Woxter, LG, Zuum

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -3989,6 +3992,10 @@ Celkon:
       model: '2GB Xpress'
     - regex: 'CliQ2(?:[);/ ]|$)'
       model: 'CliQ 2'
+    - regex: 'UniQ(?:[);/ ]|$)'
```
sgiehl, line 3995: Are you sure those detections for `UniQ` are always correct? Wondering if they can't have another android version by installing a custom rom 🤔

sgiehl (commented):
Left a comment. Otherwise looks good

regexes/device/mobiles.yml
sanchezzzhak: I didn't find any other versions of android with the above case-sensitive version

fixed it, added a version
Zoom UNIQ - android 6
Celkon UniQ  - android 7

and added new brand `New Balance`  model: NB RunIQ

sgiehl (approved):

PR #6756, opened 2021-05-29
Adds detection for Tecno Mobile devices

liviuconcioiu:
@sanchezzzhak if is good, I'll add the other devices and improve OS detection in this PR.


sanchezzzhak:
@sgiehl, @Findus23  What do you think about adding HiOS?

In addition, each manufacturer has its own OS, here is a small list.
| UI OS |  Brand | added | 
| --- | --- | --- |
| ZenUI |  Asus |  | |
| ROG UI | Asus | |
| MiUI |  Xiaomi |  |
| EMUI |  Huawei |  |
| One UI | Samsung | | 
| Oxygen OS |  OnePlus |  |
| Funtouch OS |  Vivo |  |
| HiOS |  Tecno Mobile |  |
| FIreOS |  Amazon |  ✔️  |

Findus23:
I guess if we consider Debian as an "OS" then doing the same for e.g. Osygen OS isn't that odd.

liviuconcioiu:
There is also XOS for Infinix devices. [https://en.wikipedia.org/wiki/XOS_(operating_system)](https://en.wikipedia.org/wiki/XOS_(operating_system))

sanchezzzhak:
@liviuconcioiu
in this PR you need to add only devices without HiOS.
it can be done, but it needs to be done in a different way, but first of all we need to discuss it.

I am in favor of adding more OS, but the definition by device models in the current oss.yml structure is not suitable.
when it's one record it's okay, and when it's 15-30 it's already a lot.

sgiehl:
Actually I think we should add a new OS only if it can be detected with the useragent without looking at the device model. So if there is no specific fragment or fragment combinations that explicitly point to a specific OS we shouldn't add it.
In this case I don't think the detection is unambiguous. If someone installs a vanilla Android or any custom rom on such a device, the useragent might still be the same, but the OS actually would have changed. Right? Also I wonder why the OS version is directly related to a specific device, doesn't this vendor provide updates for their os? 🤔

sanchezzzhak:
this is a sample list.
If there is a regression of android version and a particular device in a smaller direction, it is illogical
| Android version | HiOS |
| --- | --- |
| 6.1 | 2 |
| 7 | 3 |
| 8 | 4 |
| 8.1 | 4.1 |
| 9 | 5 |
| 10 | 6 |
| 11 | 7 |
the same list can be made for Infinix Mobile

sanchezzzhak (approved):

PR #6763, opened 2021-06-02
Detect new brand: Xtratech, Tambo, XGIMI, AOYODKG, Raspberry and Detect devices for existing brands: Kodak, OnePlus, Sky, Hometech, Mobicel, Maxwest, Turbo-X, Vastking, Vonino, Huawei

sanchezzzhak:
this PR merge after #6758


regexes/device/mobiles.yml
```diff
@@ -16064,11 +16080,15 @@ UTOK:
   device: 'smartphone'
   model: '$1'
 
-# Vastking
+# Vasting
```
sgiehl, line 16083: was that change on purpose? wondering why the name wasn't changed as well

sgiehl (approved):

PR #6765, opened 2021-06-04
Adds Improved Roku detection and detect app Netflix

### LINKED ISSUE #6764: Roku device detection ?
I,  I can see a lot of useragent like this "Mozilla/5.0 (QSP; Roku; AP; 6.2.80.2752)" that end up not being detected to a specific device family.

I would think that this is either a Roku TV or a Roku Stick.

Is this a known issue ? Is there a way to actually return a device family for these user agents ?

sanchezzzhak:


sanchezzzhak:
@sgiehl -  tests began to fall for some unknown reason on the latest version of macOS recently the operating system received an update (May 24, 2021)

sgiehl:
@sanchezzzhak we could try using `macos-10.15` instead of `macos-latest` to avoid such issues. Or we ignore it till it's fixed. Can disable the requirement for the macos builds temporarily. What would you suggest to do?

sanchezzzhak:
@sgiehl
let's try the version below, and in a couple of weeks we'll try to restore macos-latest.
I think pcre2 will be fixed by this time

sgiehl:
Ok. Feel free to create a PR to change that.

sanchezzzhak:
@sgiehl switching the OS version to an older version didn't help.

the error occurs in all versions of php 7.4.20, 8.0.6, 8.0.7.
if we disable pcre.jit, then everything will work. #6770

sgiehl (approved):

PR #6770, opened 2021-06-05
git-action disable pcre.jit for macos

sanchezzzhak:


sgiehl (dismissed):
Thanks for the quick fix. Maybe we can revert this PR in a couple of week if mac os works again....

.github/workflows/phpunit.yml
```diff
@@ -34,3 +34,28 @@ jobs:
         run: |
           php -v
           ./vendor/bin/phpunit
+
+  build2:
+    name: PHPUnit
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      matrix:
+        operating-system: [macos-latest]
```
sgiehl, line 43: Seems we need to keep the case, so the build name stays the same

```suggestion
        operating-system: [macOS-latest]
```

sgiehl (approved):

PR #6791, opened 2021-07-03
Add Statamic port

liviuconcioiu:


sgiehl:
I guess that is less of a port to another language but more some kind of integration into another software.
I actually wouldn't add it to the list. But we could consider adding a new list with integrations into other tools. But there actually might be a plenty of it like https://packagist.org/packages/dungeonworx/laravel-devicedetector, https://packagist.org/packages/crossknowledge/devicedetect-bundle, https://packagist.org/packages/acsiomatic/device-detector-bundle, https://packagist.org/packages/dotkernel/dot-user-agent-sniffer, ...

sanchezzzhak:
@sgiehl what about my package? https://www.npmjs.com/package/node-device-detector.
I have been supporting it for a long time (several years), since it is our main detector in nodejs projects.

on the PR.
the current list should contain only ports for other programming languages

sgiehl:
@sanchezzzhak Feel free to add your NodeJS port to the list.

PR #6792, opened 2021-07-05
Detect apps: YakYak, Teams, Raindrop.io, Mailspring, Franz, Basecamp, Notion, Streamlabs OBS, Evernote, Vuhuv, OfferUp, Rambox Pro, TuneIn Radio, Strimio, Blitz and Fix detect SuperBird browser

sanchezzzhak:


Tests/Parser/Client/fixtures/mobile_app.yml
```diff
@@ -581,3 +581,9 @@
     type: mobile app
     name: Netflix
     version: ""
+-
+  user_agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) yakyak/1.5.0 Chrome/58.0.3029.110 Electron/1.7.10 Safari/537.36
+  client:
+    type: mobile app
```
sanchezzzhak, line 587: I do not agree with the type of client, since it is desktop, what type of parser applications can this be attributed to?

sgiehl: I guess that type is a bit inaccurate. Should maybe be named only `Application`. In many cases it's not possible to say if an app is used mobile or on desktop as they have similar useragents in that case.
At some point we maybe could consider to rename that type. But that will require a new major release as it would break bc. Maybe for now it's enough to document somewhere that `mobile app` includes any type of application

sanchezzzhak: I think we need to add the ability to override the type in the yml file itself, and also add the client type "app" or semantically "desktop app".
```
# Blitz (https://blitz.gg/)
- regex: 'Blitz/([\d\.]+)'
  name: 'Blitz'
  type: 'desktop app'
  version: '$1'
```

sgiehl: I actually wouldn't distinguish between mobile and desktop in this case at all. There is actually no need. Knowing if it's an `App` is enough as you can get the device type in another way.

sanchezzzhak: :ok_hand: I'll do it in the next PR

sgiehl: Maybe wait with renaming it to `app` for a while. As it breaks backwards compatibility this needs to be done in a new major release.
Once I'm done with the reviews (hope I will find some more time this week), I would release a new minor version.

sgiehl: Let's do that change in a new PR. Would you mind solving the conflicts?

sgiehl (dismissed):

Tests/Parser/Client/fixtures/mobile_app.yml
sanchezzzhak: ok, I will resolve the conflicts in a couple of minutes

sgiehl (approved):

PR #6805, opened 2021-07-21
Adds detection for Gobuster bot/tool

sgiehl:


sanchezzzhak (approved):
There is a similar project that was added as a bot. 
https://github.com/robertdavidgraham/masscan
the actions of both repositories are similar, - this is scans

PR #6837, opened 2021-09-02
Detect devices for existing brands: Alcatel, MTC, HTC, Nokia, Vivo, Samsung, OnePlus, Nomi, ONN, Tecno Mobile, Realme, Blackview, LG, Lenovo, Motorola, Digma, Fairphone, Homtom

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -334,35 +337,24 @@ Nokia:
...
     - regex: 'Nokia 3\.1 Plus'
       model: '3.1 Plus'
-    - regex: 'Nokia 3\.1'
-      model: '3.1'
-    - regex: 'Nokia 2\.1'
-      model: '2.1'
     - regex: 'Nokia 1 Plus'
       model: '1 Plus'
-    - regex: 'Nokia 1'
-      model: '1'
```
sgiehl, line 363: Were those detection incorrect? Couldn't see a simpler replacement 🤔

sanchezzzhak: these are redundant rules, and some rules such as `Nokia 1`
blocks the detect of models 1.3 and 1.4

https://github.com/matomo-org/device-detector/blob/9e931648250981b357d78c8183da99ce840a666a/regexes/device/mobiles.yml#L452-L453

Should I restore the changes back?

sgiehl: Ah. I had searched for it, but guess I missed the other rule. No need to revert anything. It's fine to remove redundant rules.

sgiehl (approved):

PR #6850, opened 2021-09-14
OS family added to operating system parse result

Khutorny:
### Description:
Title: OS family added to Operating system parse result

Problem:
There are some use cases when operating system family is needed. Actual DeviceDetector class has no ability to get operating system family directly. I this case I need to inject OperatingSystem parser to my services and it is not good.

Resolution:
- family key with operating system family added to the result array of OperatinSystem->parse method
- operating system families mapping moved to OperationSystem parser due to its responsibility
- isDesktopOs methos added to OperatingSystem parser and used in DeviceDetector
- family key added to all os tests
- all tests passing correctly

### Review


Findus23:
I don't have a strong opinion either way, but wouldn't it be good to have this consistent with the browser family.

This is how I fetch the data on https://devicedetector.lw1.at/

https://github.com/Findus23/devicedetector.net/blob/c4ba72bb401504b4980d3e0d8f7eef4c414dbf97/public/ua.php#L42-L48

Khutorny:
I need to use `OperatingSystem::getOsFamily($dd->getOs('short_name'))` construction (the same is for browser family)

I'm writing facade for device detector in my project as Symfony bundle and have construction like:
```
class DeviceDetectorFacade
{
    private DeviceDetector $deviceDetector;

    public function __construct(DeviceDetector $deviceDetector)
    {
        $this->deviceDetector = $deviceDetector;
    }

    public function getUserAgentInfo(string $userAgent): UserAgentInfo
    {
        // use DeviceDetector
    }
}
```
So I need to inject or just statically call OperationSystem::getOsFamily
That is not good because OperatingSystem class is just a parser but not a component service so I don't want to inject it. More of that I don't want to call it statically to have hidden dependency

PR's target to save DeviceDetector class as only entrypoint to the component.
The same refactoring may be done for BrowserFamily

sanchezzzhak (dismissed):
I'm not against these changes
I myself also output os.family and browser.family in node-device-detector

https://iehol.sse.codesandbox.io/#Mozilla/5.0%20(X11;%20Linux%20x86_64)%20AppleWebKit/537.36%20(KHTML,%20like%20Gecko)%20Chrome/93.0.4577.63%20Safari/537.36

these changes will not affect the overall working code either in matomo or anywhere else.

@sgiehl merge at your discretion

sgiehl:
@Khutorny Thanks for creating the PR. Would you mind catching up @Findus23 suggestion and apply similar changes to the browser parser? So `family` is included in the default result so this is kind of consistent?

Khutorny:
Of course I can make changes for Browser parser too.
Just tell me to do that in this PR as next commit or in separate PR

sgiehl:
It's fine to simply add it to this PR

Khutorny:
@sgiehl @sanchezzzhak 
I've added browser family to Browser parse result.
Also I've sorted browsers in browser families list alphabetically for easily searching
All tests passed.

But I haven't added family to tests fixtures.
So there is a construction in BrowserTest like
```
unset($browser['short_name']);
unset($browser['family']);
```

I don't like that, but for now left as is.

If you like I can add new browser_full.yaml file with full response structure for some user agents and write separate test or test method for that case without unsets and deprecate current realization. In future it is possible to migrate all user agents to full scheme but there is a lot of 'monkey work' to add short_name and family to almost 500 cases

sanchezzzhak:
> If you like I can add new browser_full.yaml file with full response structure for some user agents and write separate test or test method for that case without unsets and deprecate current realization. In future it is possible to migrate all user agents to full scheme but there is a lot of 'monkey work' to add short_name and family to almost 500 cases

to the current file `device-detector/Tests/Parser/Client/fixtures/browser.yml `, to save the editing history 
as it was done in oss. yml

`short_name` does not need to be added.
we are slowly getting rid of this property, in the future when it is not needed at all, it will be deleted.

Khutorny:
Added browser family to browser.yaml cases

sanchezzzhak (approved):

PR #6857, opened 2021-09-17
Rename brand 2F Mobile to F2 Mobile

sanchezzzhak:
2F Mobile and F2 Mobile these are the same brands
https://storage.googleapis.com/play_public/supported_devices.html


sgiehl (approved):
As F2 mobile had been added recently, guess it should be fine to remove it again, as it hadn't been included in any release.

PR #6863, opened 2021-09-21
Detect browser GoBrowser for Symbian

sanchezzzhak:
the last activity in network traffic was in 2020
<s>
there is also such a user agent
`Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.3.3029.49 GoBrowser/58.3.3029.49 MRCHROME SOC Safari/537.36`

but it's like a bot</s>

### DIFF
--- a/Tests/Parser/Client/fixtures/browser.yml
+++ b/Tests/Parser/Client/fixtures/browser.yml
@@ -3914,3 +3914,13 @@
     engine: Blink
     engine_version: ""
     family: Chrome
+
+-
+  user_agent: Nokia6120c/GoBrowser/2.0.290
+  client:
+    type: browser
+    name: GoBrowser
+    version: 2.0.290
+    engine: ""
+    engine_version: ""
+    family: Unknown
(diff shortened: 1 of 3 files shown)


Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -3914,3 +3914,13 @@
...
+
+-
+  user_agent: Nokia6120c/GoBrowser/2.0.290
+  client:
+    type: browser
+    name: GoBrowser
+    version: 2.0.290
+    engine: ""
+    engine_version: ""
+    family: Unknown
```
sgiehl, line 3926:
```suggestion
    family:
```

sanchezzzhak: in nodejs yaml parser result
`family: ` result `family: null`
I suggest making an explicitly empty string
`family: ""`

sgiehl (approved):

PR #6902, opened 2021-11-09
Improves Android, iPadOS, iOS, macOS version detection

liviuconcioiu:


regexes/oss.yml
```diff
@@ -184,6 +184,10 @@
   name: 'Android'
   version: '$1'
 
+- regex: 'A/(\d+[\.\d]*)/'
```
sanchezzzhak, line 187: we need a more strict rule
`^A/(\d+[\.\d]*)/`

sanchezzzhak (approved):

PR #6915, opened 2021-11-20
Adds detection for Huawei Browser Mobile, iBrowser, iBrowser Mini and improves detection for other browsers

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -787,6 +787,11 @@
   name: 'Huawei Browser'
   version: '$1'
 
+# Huawei Browser for PC
+- regex: 'HBPC/(\d+[\.\d]+)'
+  name: 'Huawei Browser for PC'
+  version: '$1'
```
sanchezzzhak, line 793: set default engine blink

regexes/client/browsers.yml
```diff
@@ -787,6 +787,11 @@
   name: 'Huawei Browser'
   version: '$1'
 
+# Huawei Browser for PC
+- regex: 'HBPC/(\d+[\.\d]+)'
+  name: 'Huawei Browser for PC'
```
sanchezzzhak, line 792: @sgiehl need help. I'm not sure about the name, usually we try to avoid entering the name for browsers with verbs
Examples: Edge, Chrome, Chrome Mobile, Qwant Mobile.

what about the name -   `Huawei Browser Desktop`  or the current name good?

regexes/client/browsers.yml
```diff
@@ -787,6 +787,11 @@
   name: 'Huawei Browser'
   version: '$1'
```
sanchezzzhak, line 788: set default engine blink

sanchezzzhak (changes requested):

regexes/client/browsers.yml
```diff
@@ -1722,6 +1739,20 @@
...
+  version: '$1'
+
+# iBrowser (https://www.goibibo.com/)
+- regex: 'iBrowser/(\d+\.\d+)'
+  name: 'iBrowser'
+  version: '$1'
+
+- regex: 'iBrowser/'
+  name: 'iBrowser'
+  version: ''
```
sanchezzzhak, line 1754: merge to single regex
```yml 
- regex: 'iBrowser/(\d+\.[.\d]+)?'
```

sanchezzzhak (changes requested):

regexes/client/browsers.yml
sgiehl: Is the other detected `Huawei Browser` a mobile only one? If so we could rename the other to `Huawei Browser Mobile` and the desktop one simply `Huawei Browser`. That way it would match the other namings

regexes/client/browsers.yml
```diff
@@ -1722,6 +1739,16 @@
   engine:
     default: 'WebKit'
 
+# iBrowser Mini (https://www.goibibo.com/)
+- regex: 'iBrowser/Mini(\d+\.\d+)'
+  name: 'iBrowser Mini'
+  version: '$1'
+
+# iBrowser (https://www.goibibo.com/)
```
sanchezzzhak, line 1747: https://www.goibibo.com/
it looks like this site no longer applies to this browser, on the flight ticket sales page
I suggest removing the link to the site.

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #6925, opened 2021-12-07
Compatibility with PHP 8.1

sgiehl:


Parser/AbstractParser.php
```diff
@@ -329,23 +335,21 @@ protected function preMatchOverall(): ?array
     {
         $regexes = $this->getRegexes();
 
-        static $overAllMatch;
```
sgiehl, line 332: This doesn't work the same on PHP 8.1 anymore, as inherited methods now share the same static variables.

sanchezzzhak: https://3v4l.org/itePd

it looks like we'd better replace this with private properties.

sgiehl: The PR replaces it with a protected class property. Tests seem to pass again. So might be the only issue on PHP 8.1

sanchezzzhak: apparently, now with the changes, the variable is overwritten every time from the cache. (if possible)
in php 8, this state was stored globally for the session and caused errors.

sanchezzzhak (dismissed):

Parser/AbstractParser.php
sanchezzzhak: https://github.com/matomo-org/device-detector/blob/5d4f66afd3e16e1d31351ab3f138b3d0fb916a6a/.github/workflows/phpunit.yml#L65
delete row (this is to enable jit for mac os)
everything should work normally

sgiehl: > apparently, now with the changes, the variable is overwritten every time from the cache. (if possible)
in php 8, this state was stored globally for the session and caused errors.

That will only be the case if a new instance of `DeviceDetector` is created. Otherwise it should reuse the local variable.

PR #6926, opened 2021-12-07
Detect new brand LEMFO and Detect devices for existing brands: DEXP, SPC, OPPO, Krüger&Matz, Alcatel, Xiaomi and Fix Meizu browser

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -11716,99 +11716,6 @@ Fourel:
     - regex: 'EASY ?SMART ?M([32])'
       model: 'Easy Smart M$1'
 
-# Meizu
```
sanchezzzhak, line 11719: moved to the end of the list

sgiehl (approved):

PR #6956, opened 2022-01-07
Detect new brands: FEONAL, Ematic, Sunmi and Detect devices for existing brands: Xolo, Motorola, Samsung, LG, Sony, IconBIT, OPPO, POCO, Lava, Huawei, TechPad, NEC, Lenovo, Asus, KDDI, ZTE, Sharp, Alcatel, DIXON, Hurricane, LeEco, HTC, Santin, Polaroid

sanchezzzhak:


phpunit.xml.dist
```diff
@@ -12,6 +12,10 @@
          stopOnFailure="false"
          bootstrap="vendor/autoload.php"
 >
+    <php>
+        <ini name="memory_limit" value="512M" />
```
sgiehl, line 16: Are our tests meanwhile consuming so much memory?

sanchezzzhak: PHPUNIT first loads everything, gets the count of tests, and only then executes all the tests. (the default value is no longer enough in 128M)
if uses
`phpunit -d memory_limit=256M`, then this setting is not required. 

The github action uses limit of no more than 3GB

Should I remove this parameter or reduce the number to 256M?

sgiehl: Ok. makes sense. No need to change it then.

sgiehl (approved):

PR #6981, opened 2022-01-28
fix: change prop public static $clientTypes to protected

### LINKED ISSUE #6980: memory leak in DeviceDetector\DeviceDetector::addClientParser()
Hi
DeviceDetector\DeviceDetector::addClientParser() method is leaking memory.
Every time  DeviceDetector\DeviceDetector object is instantiated multiple calls to addClientParser are made, and $clientTypes static property is growing indefinitely

sanchezzzhak:


maciej-tomczak-tg (approved):
this should do the trick

sgiehl (approved):
makes sense.

PR #7004, opened 2022-02-19
Adds detection for Purism brand and fix iOS detection

### LINKED ISSUE #7003: Librem phone is counted as an iOS device
## Expected Behavior
Logged visits should be the right OS.

## Current Behavior
Visit any site with Matomo installed on the Librem 5 phone.
Look in the "visits in real time" in the left of Matomo Analytics and it registers that as "iOS"??
Browser: Chrome webview 64
even though this is the default browser that comes on Librem5. It should show it is a Linux phone not "iOS".

## Possible Solution
Check for Librem/pinephone or other Linux phones as these are going to be growing in use lately.

## Steps to Reproduce (for Bugs)
Visit any site with Matomo installed on a Librem 5 phone.
Look in the "visits in real time" in the left of Matomo Analytics and see that the pages you just visited register as iOS visits.

## Context
There is no way to know just how many Librem phone users are using the site, counting as iOS would send developer effort in the wrong direction testing sites.

## Your Environment

* Matomo Version:4.7.1
* PHP Version: 7.4.28
* Server Operating System:Ubuntu
* Additionally installed plugins: N/A
* Browser: GNOME WEB 3.38.2 default browser
* Operating System: Librem PureOS

liviuconcioiu:
ref #7003


Tests/fixtures/smartphone-3.yml
```diff
@@ -13908,3 +13908,21 @@
     model: REVVL V
   os_family: Android
   browser_family: Chrome
+-
+  user_agent: Mozilla/5.0 (Librem 5, like iPhone; X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0 Safari/605.1.15 Epiphany/605.1.15
+  os:
+    name: GNU/Linux
```
sanchezzzhak, line 13914: `Pure OS`

sanchezzzhak (approved):

PR #7007, opened 2022-02-19
Detect new brands: iPEGTOP, Nexa, Hi Nova and Detect devices for several existing brands

sanchezzzhak:


Tests/fixtures/smartphone-18.yml
```diff
@@ -3471,24 +3471,6 @@
     model: V3 5G
   os_family: Android
   browser_family: Unknown
--
```
sanchezzzhak, line 3474: duplicate test

regexes/device/mobiles.yml
```diff
@@ -7867,9 +7869,17 @@ Turkcell:
       device: 'tablet'
       model: 'T Tablet'
 
+# Hi Nova (sub brand Huawei)
+Hi Nova:
+  regex: 'Hebe-BD00'
+  device: 'smartphone'
+  models:
+    - regex: 'Hebe-BD00'
+      model: '9'  # Nova 9
```
sgiehl, line 7878: Did they split of lately? Just wondering as there are a lot phones called `Nova ` in the Huawei list.

sanchezzzhak: Huawei will continue to release Nova under the Huawei brand only in China and or those countries where there are no sanctions

sanchezzzhak: I don't think we should carry over the whole `Nova`.
only those that have a new logo.

Huawei nova 9
https://phablet.jp/?p=88149
Hi Nova 9
https://phablet.jp/?p=91619
new Hi Nova site
https://xnova.com/#/goods

sgiehl: ok. sounds good

sgiehl (approved):

PR #7020, opened 2022-02-28
Add default engine for Avast Secure Browser and AVG Secure Browser

liviuconcioiu:


regexes/client/browsers.yml
```diff
@@ -626,6 +626,8 @@
 - regex: 'AVG(?:/(\d+[\.\d]+))?'
   name: 'AVG Secure Browser'
   version: '$1'
+  engine:
+    default: 'Blink'
```
sgiehl, line 630: Were there older version of those browsers that might still have used WebKit? So versions using chromium before 28?
If so we should more likely use the same as on chrome:
```
    default: 'WebKit'
    versions:
      28: 'Blink'
```

sgiehl (approved):

PR #6989, opened 2022-02-08
Introduce browser client hint detection

sgiehl:
### Description:

As discussed in some other issues it might happen in the near future that some devices or clients might freeze the useragents in favor of client hints.

This is a proposal on how we could possibly implement a client hint detection in device detector
It's not yet finished and there are still a couple of things that might need to be discussed and improved.

I did not yet had much time to think about this or implement it, so if someone has a suggestion for a better solution feel free to suggest it here.

@sanchezzzhak Any feedback on that one?

PR #7050, opened 2022-03-10
fix hints detection, because ""

Nevercold:
### Description:

Detection error fixed.
Remove the "" from the Value

https://github.com/matomo-org/device-detector/pull/6989

### Review

* [x] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)

### DIFF
--- a/ClientHints.php
+++ b/ClientHints.php
@@ -222,13 +222,13 @@ class ClientHints
                 case 'sec-ch-ua-arch':
                 case 'arch':
                 case 'architecture':
-                    $architecture = \str_replace('"', '', $value);
+                    $architecture = \trim($value, '"');
 
                     break;
                 case 'http-sec-ch-ua-bitness':
                 case 'sec-ch-ua-bitness':
                 case 'bitness':
-                    $bitness = \str_replace('"', '', $value);
+                    $bitness = \trim($value, '"');
 
                     break;
                 case 'http-sec-ch-ua-mobile':
@@ -240,25 +240,25 @@ class ClientHints
                 case 'http-sec-ch-ua-model':
                 case 'sec-ch-ua-model':
                 case 'model':
-                    $model = \str_replace('"', '', $value);
+                    $model = \trim($value, '"');
 
                     break;
                 case 'http-sec-ch-ua-full-version':
                 case 'sec-ch-ua-full-version':
                 case 'uafullversion':
-                    $uaFullVersion = \str_replace('"', '', $value);
+                    $uaFullVersion = \trim($value, '"');
 
                     break;
                 case 'http-sec-ch-ua-platform':
                 case 'sec-ch-ua-platform':
                 case 'platform':
-                    $platform = \str_replace('"', '', $value);
+                    $platform = \trim($value, '"');
 
                     break;
                 case 'http-sec-ch-ua-platform-version':
                 case 'sec-ch-ua-platform-version':
                 case 'platformversion':
-                    $platformVersion = \str_replace('"', '', $value);
+                    $platformVersion = \trim($value, '"');
 
                     break;
                 case 'brands':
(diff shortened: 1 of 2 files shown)


Nevercold:
I don't know, but can't do anything with the outputted errors from PHPUnit.

ClientHints.php
```diff
@@ -217,6 +217,10 @@ public static function factory(array $headers): ClientHints
         $fullVersionList = [];
 
         foreach ($headers as $name => $value) {
+            if (\is_string($value)) {
+                $value = \str_replace('"', '', $value);
```
sgiehl, line 221: this will actually break the detection of `sec-ch-ua-full-version-list`, as it contains a coulple of `"` on purpose. We maybe should do something like this for each expected string value:
```php
$x = preg_replace('/^"(.*)"$/', '$1', $value);
```

Nevercold: or str_replace only for platformversion, platform and uafullversion

sgiehl: yes. But I'm not sure if there might be cases in the future where a `"` is contained within as well. But guess we can also handle it if it makes problems later.

ClientHints.php
```diff
@@ -222,13 +222,13 @@ public static function factory(array $headers): ClientHints
                 case 'sec-ch-ua-arch':
                 case 'arch':
                 case 'architecture':
-                    $architecture = $value;
+                    $architecture = \str_replace('"', '', $value);
```
sgiehl, line 225: Just got another idea. We could simply use `trim` so any `"` in the middle would be untouched.
```suggestion
                    $architecture = \trim($value, '"');
```

Nevercold: sounds good.
So it should be back at the beginning for all?

sgiehl: No. That would still break `sec-ch-ua-full-version-list`, as that one requires the surrounding `"`.

Nevercold: yes, I noticed that myself. :P

README.md
```diff
@@ -44,6 +44,10 @@ use DeviceDetector\Parser\Device\AbstractDeviceParser;
 // for other options see VERSION_TRUNCATION_* constants in DeviceParserAbstract class
 AbstractDeviceParser::setVersionTruncation(AbstractDeviceParser::VERSION_TRUNCATION_NONE);
 
+// needed for clientHints
+header('Accept-CH: Sec-CH-UA-Full-Version, Sec-CH-UA-Platform, Sec-CH-UA-Platform-Version, Sec-CH-UA-Model, Sec-CH-UA-Arc', true);
+
+
```
sgiehl, line 50: We maybe should include a new section about client hints overall instead, also explaining how they work, and that they won't fully work on the first request,.. So maybe remove it again here and we'll create another PR to add some more docs on that later.

README.md
```diff
@@ -44,6 +44,10 @@ use DeviceDetector\Parser\Device\AbstractDeviceParser;
 // for other options see VERSION_TRUNCATION_* constants in DeviceParserAbstract class
 AbstractDeviceParser::setVersionTruncation(AbstractDeviceParser::VERSION_TRUNCATION_NONE);
 
+// needed for clientHints
+header('Accept-CH: Sec-CH-UA-Full-Version, Sec-CH-UA-Platform, Sec-CH-UA-Platform-Version, Sec-CH-UA-Model, Sec-CH-UA-Arc', true);
```
liviuconcioiu, line 48: There's a missing h `Sec-CH-UA-Arc` to `Sec-CH-UA-Arch`

sgiehl (approved):

nunoperalta:
Hey - Thanks very much for this.
Just realized this isn't included in 5.0.5

Any estimate date of when will the next version be released? :)
Thank you very much.

PR #7053, opened 2022-03-12
Improvement for ClientHints

sanchezzzhak:


Parser/Client/Browser.php
```diff
@@ -584,6 +624,21 @@ public function parse(): ?array
...
 
+        $family  = self::getBrowserFamily((string) $short);
+        $appHash = $this->browserHints->parse();
+
+        if (null !== $appHash && $appHash['name'] !== $name) {
+            $name    = $appHash['name'];
+            $version = '';
+            $short   = '';
+
+            if (\preg_match('~Chrome\/.+ Safari\/537.36~i', $this->userAgent)) {
```
mimmi20, line 635: Dont you need to escape the dot too to `if (\preg_match('~Chrome\/.+ Safari\/537\.36~i', $this->userAgent)) {`?

Parser/Client/Browser.php
```diff
@@ -584,6 +624,21 @@ public function parse(): ?array
             $engineVersion = $browserFromUserAgent['engine_version'];
         }
 
+        $family  = self::getBrowserFamily((string) $short);
+        $appHash = $this->browserHints->parse();
+
+        if (null !== $appHash && $appHash['name'] !== $name) {
+            $name    = $appHash['name'];
+            $version = '';
+            $short   = '';
```
sgiehl, line 633: Does that mean that now browsers might be returned that aren't defined in the list above, and such don't have a short code?

sanchezzzhak: right, but I can add short names

sgiehl (approved):

PR #7100, opened 2022-04-27
Detect new brands: Zonko, Zeeker, TJC, HOTREALS and Detect devices for several existing brands

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -29180,6 +29218,34 @@ X-AGE:
...
+  regex: 'Metal_Tablet_10(?:[);/ ]|$)'
+  device: 'tablet'
+  model: 'Metal Tablet 10'
+
+# HOTREALS
+HOTREALS:
+  regex: 'HOTREALS'
+  device: 'tablet'
+  models:
+    - regex: 'HOTREALS(?:[);/]|$)'
```
sgiehl, line 29238: Does this device not have any other details in the useragent? Wondering if other devices of that brand might cause a incorrect detection then.

sanchezzzhak: At the moment. This is only device they have.

sgiehl (approved):

PR #7116, opened 2022-05-16
Detect browser: EcommApp

soyes:
Detect the EcommApp browser.

### DIFF
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -486,6 +486,7 @@ class Browser extends AbstractClientParser
         'XV' => 'Xvast',
         'ZE' => 'Zetakey',
         'ZV' => 'Zvu',
+        'EA' => 'EcommApp',
 
         // detected browsers in older versions
         // 'IA' => 'Iceape',  => pim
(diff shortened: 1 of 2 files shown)


sgiehl:
@soyes can you provide a link, where to get that browser app?

soyes:
Hello @sgiehl,

The browser cannot be downloaded because is a webview that is used in different mobile apps whose user agent is "EcommApp".

We could provide you the name of some of our mobile apps, but the navigation in the webview is limited to the web domain associated to the app, so you would not be able to navigate outside of that domain.

We have tested the library using a desktop browser changing the user agent to "EcommApp".

Thank you.

sgiehl:
@soyes So what makes `EcommApp` a browser? Is it simply using the systems webview component or did you implement a webview by your own?

soyes:
Hello @sgiehl,

EcommApp is a brand of apps for ecommerce that use a webview inside the app with the user agent "EcommApp".

The websites shown in the webview of these apps integrate external services (such as search engines, customer support chats, etc.) that use Mobile-Detect, so we need the webview in our apps to be detected as being used in a mobile device. That way, those services will be shown as their mobile version, instead of being shown as the desktop version.

This is why in this PR we added EcommApp as a browser instead of an app, because all of our apps use the webview with the user agent "EcommApp".

Thank you.

PR #7122, opened 2022-05-18
Detect devices for existing brands 

sanchezzzhak:


Parser/Device/AbstractDeviceParser.php
```diff
@@ -81,1300 +81,1304 @@ abstract class AbstractDeviceParser extends AbstractParser
...
+        'EZ'  => 'Ezze',
+        'UF'  => 'EYU',
+        '5F'  => 'F150',
+        'F6'  => 'Facebook',
+        'FA'  => 'Fairphone',
+        'FM'  => 'Famoco',
+        '17'  => 'FarEasTone',
+        '9R'  => 'FaRao Pro',
+        'FB'  => 'Fantec',
+        'FAC' => 'Facetel',
```
sanchezzzhak, line 421: Please note that the 2-character shortcodes have ended.

Can we transfer $deviceBrands to YML file?

sgiehl: I actually don't see much benefit in moving this to a yml file. It will actually only add ann additional yml parsing, which might slow down the detection more.

Regarding three character shortcodes:
I need to check if Matomo is able to handle them properly before we can merge and release a new version. At least the database is able to store them already, but not sure if the code expects two chars somewhere. Need to investigate that. Meanwhile it might be better to focus on adding devices for already existing brands, as we can merge such PRs.

sanchezzzhak: OK, I will move new brands from this PR in new branch

sgiehl (approved):

PR #7138, opened 2022-06-09
Fix version truncation for client hints

sgiehl:
When the browser or os version is detected using client hints, the version is currently not truncated using the configured version truncation.


sanchezzzhak (approved):
I watched it yesterday.
it looks good

PR #7163, opened 2022-07-06
Improves version detection for iPadOS

liviuconcioiu:


sanchezzzhak (approved):
Mozilla/5.0 (iPhone; CPU iPhone OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/20A5283p [FBAN/FBIOS;FBAV/373.0.0.27.112;FBBV/379597152;FBDV/iPhone11,8;FBMD/iPhone;FBSN/iOS;FBSV/16.0;FBSS/2;FBID/phone;FBLC/ru_RU;FBOP/5;FBRV/0]

I think we can add this without a test.
iPhone already contains a version 16.

PR #7193, opened 2022-08-01
Optimise devices detection for apple

sanchezzzhak:
I replaced the greedy regular expressions with all the values that fail.

before the changes, the detection speed was (with the turbo fan turned on)

* Mozilla/5.0 (iPhone; CPU iPhone OS 15_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 Instagram 239.2.0.17.109 (iPhone9,3; iOS 15_5; it_IT; it-IT; scale=2.00; 750x1334; 376668393) NW/3

detector.parseDevice (deviceIndexes on) x 423 ops/sec ±1.19% (88 runs sampled)
detector.detect (indexes on) x 350 ops/sec ±0.79% (88 runs sampled)

after:
* Mozilla/5.0 (iPhone; CPU iPhone OS 15_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 Instagram 239.2.0.17.109 (iPhone9,3; iOS 15_5; it_IT; it-IT; scale=2.00; 750x1334; 376668393) NW/3

detector.parseDevice (deviceIndexes on) x 2,721 ops/sec ±2.06% (89 runs sampled)
detector.detect (indexes on) x 1,048 ops/sec ±2.95% (85 runs sampled)

on the second cycle, we have become 6.4 times faster.


sgiehl (approved):
Looks good to me.

PR #7197, opened 2022-08-04
Detect new browsers and apps - Tint Browser, hola!, RedReader

sanchezzzhak:


Parser/Client/Browser.php
```diff
@@ -68,7 +68,6 @@ class Browser extends AbstractClientParser
         'DW' => 'Dark Web Browser',
         '1B' => '115 Browser',
         '2B' => '2345 Browser',
-        '36' => '360 Phone Browser',
```
sgiehl, line 71: wondering if it wouldn't be better to keep the record and maybe give it the same name as below, so the short code is still valid and usable.

sanchezzzhak: I returned the changes.

sgiehl (approved):

PR #7217, opened 2022-08-22
Adds detection for Amerigo, Internet Browser Private, 18+ Privacy, Beyond Private, Black Lion

liviuconcioiu:


regexes/client/hints/browsers.yml
```diff
@@ -142,3 +142,8 @@
 'com.skyfire.browser': 'Skyfire'
 'com.sonymobile.smallbrowser': 'Sony Small Browser'
 'org.hola': 'hola! Browser'
+'it.ideasolutions.amerigo': 'Amerigo'
+'org.xbrowser.prosuperfast': 'Internet Browser Private'
```
sanchezzzhak, line 146: I think it would be better if we use an alternative name
`xBrowser Pro Super Fast`

sanchezzzhak (approved):

PR #7222, opened 2022-08-23
Adds detection for various browsers

liviuconcioiu:


liviuconcioiu:
@sanchezzzhak you have to check those two browsers from your list, since I don't have writing permission.

```
com.ume.browser.northamerica
com.ume.browser.newage
```

sanchezzzhak (approved):

PR #7228, opened 2022-08-30
Detect devices for existing brands and new brands

sanchezzzhak:


sgiehl (changes requested):

regexes/device/mobiles.yml
```diff
@@ -5019,7 +5024,7 @@ Casper:
 
 # Cat (Caterpillar, Catphones)
 Cat:
-  regex: 'Cat ?(tablet|stargate|nova)|(?:B15Q|CAT B35|CAT-B100|S62 Pro|S48c)(?:[);/ ]|$)'
+  regex: 'Cat ?(tablet|stargate|nova)|(?:B15Q|CAT B35|CAT-B100|S62 Pro|S48c|Android 9.*; S52|Android 10; S42)(?:[);/ ]|$)'
```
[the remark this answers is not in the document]
sanchezzzhak: after Android <version> there is a language construct
exsample
Android 9; en; Model code) applewebkit/
Android 9; en; Model code build/<os build>) applewebkit/

the  separator ";", present only in parentheses.

as for the Cat S52, it exists only since version 9 of android 
https://www.gsmarena.com/cat_s52-9938.php

The s52 code can also occur for Doogee, on Android version 12, and I don't have such a user agent.

image

https://www.imei.info/phonedatabase/sfr-altice-s42/
https://www.imei.info/phonedatabase/sfr-altice-s32/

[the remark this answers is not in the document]
sgiehl: Ok. I see. But then you could also use something like `Android 9;( \w+;)? S52`

regexes/device/mobiles.yml
```diff
@@ -30,7 +30,7 @@ Ace:
 
 # SFR or SFR Mobile
 SFR:
-  regex: 'StarShine|StarTrail|STARADDICT|StarText|StarNaute|StarXtrem|StarTab|(?:Android 8.*; (?:AURA (?:PLUS|ULTRA)|AURA))(?: Build|[;)])'
+  regex: 'StarShine|StarTrail|STARADDICT|StarText|StarNaute|StarXtrem|StarTab|(?:Android 8(?:[\d.]*);(?: [\w-]+;)? (?:AURA (?:PLUS|ULTRA)|AURA))(?: Build|[;)])'
```
sgiehl, line 33: If you want to match a dot, you need to escape it. Otherwise it's a placeholder for any random character.
```suggestion
  regex: 'StarShine|StarTrail|STARADDICT|StarText|StarNaute|StarXtrem|StarTab|(?:Android 8(?:[\d\.]*);(?: [\w-]+;)? (?:AURA (?:PLUS|ULTRA)|AURA))(?: Build|[;)])'
```
same might apply to the other changes.

sanchezzzhak: there is no need to escape the dot in square brackets, this is not any character.

here is a demo

sgiehl: You're right. Should have taken a closer look and think about it, instead of trying to do a quick review while doing something else 🙈

sgiehl (approved):

PR #7235, opened 2022-09-10
Detect devices for existing brands and new brands

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -15297,10 +15307,10 @@ Nvidia:
 
 # O+ (OPlus)
 O+:
-  regex: 'O\+[ _](Air|Crunch|Ultra|Upsized|Venti|360 ALPHA PLUS 2.0)'
+  regex: 'O\+[ _](8.38z|8.97|Air|Crunch|Ultra|Upsized|Venti|360 ALPHA PLUS 2.0)'
```
sgiehl, line 15310: Shouldn't here the dots be escaped to prevent a match for any character?
```suggestion
  regex: 'O\+[ _](8\.38z|8\.97|Air|Crunch|Ultra|Upsized|Venti|360 ALPHA PLUS 2.0)'
```

regexes/device/mobiles.yml
```diff
@@ -15297,10 +15307,10 @@ Nvidia:
...
   device: 'smartphone'
   models:
-    - regex: 'O\+[ _](Crunch|Ultra 2.0|Ultra|Upsized|Venti 4G)'
+    - regex: 'O\+[ _](Crunch|Ultra 2.0|Ultra|Upsized|Venti 4G|8.38z|8.97)'
```
sgiehl, line 15313:
```suggestion
    - regex: 'O\+[ _](Crunch|Ultra 2.0|Ultra|Upsized|Venti 4G|8\.38z|8\.97)'
```

sgiehl (changes requested):

sgiehl (approved):

PR #7265, opened 2022-10-14
Fix issue when providing an array with numeric index for client hint factory

sgiehl:
### Description:

The factory for client hints might throw an error on PHP 8, if the provided array contains a record with a numeric key, as `strtolower` doesn't accept integers anymore.

### Review

### DIFF
--- a/ClientHints.php
+++ b/ClientHints.php
@@ -236,7 +236,7 @@ class ClientHints
         $fullVersionList = [];
 
         foreach ($headers as $name => $value) {
-            switch (\str_replace('_', '-', \strtolower($name))) {
+            switch (\str_replace('_', '-', \strtolower(\strval($name)))) {
                 case 'http-sec-ch-ua-arch':
                 case 'sec-ch-ua-arch':
                 case 'arch':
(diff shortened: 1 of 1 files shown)


ClientHints.php
```diff
@@ -236,7 +236,7 @@ public static function factory(array $headers): ClientHints
         $fullVersionList = [];
 
         foreach ($headers as $name => $value) {
-            switch (\str_replace('_', '-', \strtolower($name))) {
+            switch (\str_replace('_', '-', \strtolower(\strval($name)))) {
```
sanchezzzhak, line 239: I would replace \strval($name) with (string)$name
> [EA] '(string) $language' can be used instead (reduces cognitive load, up to 6x times faster in PHP 5.x). 
 Inspection info: Analyzes if PHP4 functions (intval, floatval, doubleval, strval) are used for type casting and generates hints to use PHP5's type casting construction (i.e. '(type) parameter').

sgiehl: Not sure if that still makes a difference in PHP 7 or 8. But will change it.

sanchezzzhak (approved):

PR #7271, opened 2022-10-18
Adds client hints test for Brave browser

liviuconcioiu:


sanchezzzhak (approved):

Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -7141,3 +7141,16 @@
   headers:
     Sec-CH-UA: '"Chromium";v="92", " Not A;Brand";v="99", "HuaweiBrowser";v="92"'
     Sec-CH-UA-Mobile: '?1'
+-
+  user_agent: Mozilla/5.0 (Linux; Android 10) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Mobile Safari/537.36
+  client:
+    type: browser
+    name: Brave
+    version: "106"
```
sanchezzzhak, line 7149: should we use the full version of the browser from the user agent in the future?

liviuconcioiu: That's what I think. Because the versions aren't right. I've updated PR #7065.

PR #7284, opened 2022-11-04
Adds detection for various libraries and improve detection for iPadOS

liviuconcioiu:


regexes/client/libraries.yml
```diff
@@ -5,6 +5,21 @@
...
+  name: 'got'
+  version: ''
+
+# Typhoeus (https://github.com/typhoeus/typhoeus)
+- regex: 'Typhoeus'
+  name: 'Typhoeus'
+  version: ''
+
+# ReqBin (https://reqbin.com/curl)
+- regex: 'ReqBin Curl Client/([\.\d]+)'
```
sanchezzzhak, line 19: I have not found a library that gives out such a user-agent.
I propose to transfer this definition of UA to bots.

sanchezzzhak (dismissed):

regexes/client/libraries.yml
```diff
@@ -357,3 +367,8 @@
 - regex: 'harbor-registry-client'
   name: 'Harbor registry client'
   version: ''
+
+# Generic detections
+- regex: 'github'
```
sanchezzzhak, line 372: @sgiehl can you look at this condition is it acceptable to enter a common name for undefined libraries into the detector?

sgiehl: I wouldn't add generic libraries. They might also be hard to differ from a bot :thinking:

sanchezzzhak (approved):

PR #7301, opened 2022-12-04
Add Axios library

XAwaludin:


sanchezzzhak (approved):
accepted, on the grounds that there is such a library used in node js/deno or web

Issue #7340, opened 2023-01-25
WebKitGTK-based web browsers (on Linux) are considered to be "Safari" in Matomo reports


nekohayo:
## Current Behavior

When visiting a Matomo-tracked website using [GNOME Web](https://wiki.gnome.org/Apps/Web/) (a.k.a. [Epiphany](https://en.wikipedia.org/wiki/GNOME_Web)), Matomo interprets it as being "Safari on Linux", which doesn't exist, strictly speaking:

![2023-01-25 epiphany as seen by Matomo](https://user-images.githubusercontent.com/479401/214632520-dc95a39a-e670-4251-88a6-4ab8ab2dfd82.png)

Note that Epiphany is using WebKitGTK, which is not strictly the same as WebKit vanilla.

## Expected Behavior

Ideally, Matomo should find a way to detect which webkit-based Linux browser the user is using, or at least report it as WebKit on Linux or something separate than Safari.

## Possible Solution

I know that user agent strings are an [unresolved mess of epic proportions](https://webaim.org/blog/user-agent-string-history/) in the industry, but I don't know what else to suggest; nothing else has been adopted yet, as far as I've been told.

Visiting some user agent tester website with Epiphany "technology preview" (the dev version that will become Epiphany / GNOME Web 44), I can see this being the user agent:

`Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.0 Safari/605.1.15`

Essentially, I think the only differentiator is that it's on "Linux" rather than on macOS or iOS (and even then, in mobile versions it has to pretend like it's Safari on iOS...). I am told that "Adding application branding to the user agent header is not web compatible; notably, Epiphany does not do so anymore", so you probably have to do the interpretation on the Matomo side... at least, we can't really call it Safari.

## Your Environment

* Matomo Version: 4.13.1
* PHP Version: 7.4.33
* Server Operating System: CentOS Stream 8
* Additionally installed plugins: none
* Browser: GNOME Web
* Operating System: Fedora 37 (Linux)

mcatanzaro:
We don't support client hints, but if we did, we would use them to lie. From WebKit's perspective, we don't really want to make device detection easy.

sgiehl:
In that case, let me simply close this issue 🙈

PR #7341, opened 2023-01-26
Detect GNOME Web if is this linux OS and browser safari

sanchezzzhak:
#7340 
I agree that Safari cannot be in linux, so this browser has been replaced with GNOME Web, in this fix.

### DIFF
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -939,7 +939,7 @@ class Browser extends AbstractClientParser
             $engine        = $this->buildEngine($regex['engine'] ?? [], $version);
             $engineVersion = $this->buildEngineVersion($engine);
 
-            if ('Safari' === $name && $this->matchUserAgent('X11; Linux')) {
+            if ('Safari' === $name && $this->matchUserAgent('\(X11; Linux')) {
                 $name         = 'GNOME Web';
                 $version      = '';
                 $browserShort = 'EP';
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -939,6 +939,12 @@ class Browser extends AbstractClientParser
             $engine        = $this->buildEngine($regex['engine'] ?? [], $version);
             $engineVersion = $this->buildEngineVersion($engine);
 
+            if ('Safari' === $name && $this->matchUserAgent('X11; Linux')) {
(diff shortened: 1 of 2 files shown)


Exalm:
Well, is it's also not always GNOME Web, it can also be any of the many other browsers using WebKitGTK or WPE WebKit - e.g. Eolie, Ephemeral, Tangram, Midori, etc, as well as webkit views in apps like Builder or Feeds.

sanchezzzhak:
Can we come up with a common name for this?
what about the name: 
*  Webkit-based browser
*  WebKitGTK
*  WebKit View

your variants?

Exalm:
I'd say WebKit, but really a question for @mcatanzaro

mcatanzaro:
Um, is Matomo something that's widely used by various websites for browser detection? Because if so, detecting anything other than Safari would probably threaten our continued existence.

If it's just for analytics purposes, then it's fine to return the more accurate result "WebKit." But if websites would look at this detection result to decide to alter their web content based on the browser they detect, then we _really_ need to be detected as Safari.

Simbiat:
AFAIK, Safari detection generally is required on front-end, when you are dealing with JS code, since Safari does not support quite  a few features, that Webkit browsers do. For those detection is done on frontend side, though, in JS code, not on PHP backend. I have not seen or heard about instances, where you detected browser on backend and changed something in the HTML code generated from it. But... Theoretically it could be the case. For example if the mentioned JS code is generated by a PHP script. Something in me pains at the thought of such a use case, but you cannot exclude it.
Perhaps, have some sort of a setting which will determine whether to show Safari or WebKit?

Exalm:
The problem is a lot of fairly popular websites are like: if this is not safari, chrome or firefox (or edge or a few more popular ones) - show "this browser is not supported" instead of actually checking available features. WhatsApp Web would be an example. So you can support everything safari supports and will still get denied because the websites don't know other browsers exist.

mcatanzaro:
> I have not seen or heard about instances, where you detected browser on backend and changed something in the HTML code generated from it.

It happens quite a lot, unfortunately.

Analystics is fine, but if your device-detector can be used by websites to alter web content, then it's best for us if you just close this issue and forget it ever existed. ;)

PR #7414, opened 2023-05-29
fix Android tablet detection in VERSION_TRUNCATION_MAJOR mode

blinkseb:
### Description:

Hello,

First, thanks a lot for this awesome library!

Because of privacy concerns, I'm using this library in `VERSION_TRUNCATION_MAJOR`, and I noticed some issues when parsing some old Android 3 tablet user-agents. The device ends up empty instead of `tablet`.

Looking at the code, it's because of the builtin `version_compare`, for which, I'm not sure why, `3` is lower than `3.0` instead of being equals. I fixed the issue by using only major versions in `version_compare`, instead of major + minor.

For the tests, I've just duplicated the existing ones, running with `VERSION_TRUNCATION_MAJOR` instead of `VERSION_TRUNCATION_NONE`. There's probably a better / faster solution, please tell me what you think.

Thanks a lot in advance!

### Review

### DIFF
--- a/DeviceDetector.php
+++ b/DeviceDetector.php
@@ -980,10 +980,10 @@ class DeviceDetector
          * Devices running Android 3.X are tablets. Device type of Android 2.X and 4.X+ are unknown
          */
         if (null === $this->device && 'Android' === $osName && '' !== $osVersion) {
-            if (-1 === \version_compare($osVersion, '2.0')) {
+            if (-1 === \version_compare($osVersion, '2')) {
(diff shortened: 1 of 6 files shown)


blinkseb:
It was a bit harder than expected, because the browser version is used in regexes to guess the browser engine, and, when truncated, it leads to different results.

I extracted the version truncation from the `buildVersion` to a dedicated `truncateVersion` and updated the code accordingly. This allows to use the full version to guess the engine, while propagating the truncated version.

Let me know what you think.

Thanks!

PR #7430, opened 2023-06-30
Detect brand and model device for Client-Hints

sanchezzzhak:
these statistics are where the device model is known and the brand is not known from the client-hints.

|year|month|sum(device_model_id  > 0 and device_brand_id = 0)|
|----|-----|-----------------------------------------------------------------|
|2023|4|10056|
|2023|5|69586|
|2023|6|104428|
|2023|7|8652|

image

this fix will help to return the brand and the name of the device.


Parser/Device/AbstractDeviceParser.php
```diff
@@ -1861,6 +1861,16 @@ public function parse(): ?array
...
         $deviceModel      = $resultClientHint['model'] ?? '';
 
+        // is freeze user-agent then restoring the original UA for the device definition
+        if ('' !== $deviceModel && \preg_match('~Android 10[.\d]*; K(?: Build/|[;)])~i', $this->userAgent)) {
+            $osVersion = $this->clientHints->getOperatingSystemVersion();
+            $this->setUserAgent(\preg_replace(
+                '(Android 10[.\d]*; K)',
+                \sprintf('Android %s; %s', $osVersion, $deviceModel),
+                $this->userAgent
+            ));
```
sgiehl, line 1871: In theory it could happen that a device model was provided through the client hints, but the os version is empty. This would in that case end up in a user agent where no version is set at all. Maybe we should keep Android 10 if no version is provided?

sanchezzzhak: We have rules based on the version + code of the device. I will add a check for the presence of OS

sanchezzzhak: @sgiehl You can watch this PR again? I made changes.

sgiehl (approved):

PR #7490, opened 2023-10-21
Harmonizing `feed reader` type

Simbiat:
### Description:

Parser name for feed readers is `feed reader`, but we had apps with type `Feed Reader` or even `Feed Reader App`, which was inconsistent compared to other client types both in terms of case and name. This little change harmonizes things.

### Review

* [+] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [+] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [+] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [+] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [+] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [+] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [+] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [+] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [+] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [+] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [+] Existing documentation updated if needed


Simbiat:
Can the checks be rerun? It does not look like they failed because of my change, after all.

liviuconcioiu:
> Can the checks be rerun? It does not look like they failed because of my change, after all.

If you use Github Desktop, it can be done from the app.

Untitled1

Simbiat:
I committed through GitHub website. Cloned the repo in GitHub Desktop, triggered recheck there, and it shows as if the check were scheduled, but I do not see any updates on website.

liviuconcioiu:
You can make a test commit and remove it in another commit then push both. It will trigger the checks.

Simbiat:
Done. Probably need to report a bug to GitHub, too.

sgiehl:
Doesn't this change make the `type` flag for feed readers fully obsolete? It would be the same everywhere. So might be better to remove it from the yml file (and maybe set it fixed in PHP if needed)

Simbiat:
Yup, no need to keep them. By default we are using `$parserName` value for this from the looks of it, so only need to modify YAML file.

Simbiat:
Test was failing on YAML structure check, not the parser check. Updated the test accordingly.

sanchezzzhak (approved):

PR #7498, opened 2023-11-07
Adds detection for various libraries, bots and operating systems

liviuconcioiu:


regexes/oss.yml
```diff
@@ -555,7 +569,7 @@
   name: 'CentOS Stream'
   version: '$1'
 
-- regex: '.+.el(\d+(?:[_\.]\d+)*).(?:centos|x86_64)'
+- regex: '.+.el(\d+(?:[_\.]\d+)*)'
```
sanchezzzhak, line 572: we need a more strict rule. what about?
`Linux/(?:\d+\.[\d.-]+)\.el(\d+(?:[_.]\d+)*)`

regexes/oss.yml
```diff
@@ -5,6 +5,20 @@
 # @license http://www.gnu.org/licenses/lgpl.html LGPL v3 or later
 ###############
 
+##########
+# Oracle Linux (https://www.oracle.com/linux/)
+##########
+- regex: '.+.el(\d+(?:[_\.]\d+)*)uek'
```
sanchezzzhak, line 11: we need a more strict rule. what about?
`Linux/(?:\d+\.[\d.-]+)\.el(\d+(?:[_.]\d+)*)uek` 

https://regex101.com/r/xs0M2a/1

regexes/bots.yml
```diff
@@ -3472,6 +3472,35 @@
   category: 'Service Agent'
   url: 'https://www.phpmyadmin.net/'
 
+- regex: 'Matomo/([\d+.]+)'
```
sanchezzzhak, line 3475: @sgiehl will there be any conflicts in motamo itself, if we add this?

regexes/oss.yml
liviuconcioiu: @sanchezzzhak I've reverted the commits for CentOS, since I'm not sure if this user agent is for Red Hat Enterprise Linux or CentOS

`python-requests/2.7.0 CPython/2.7.3 Linux/2.6.18-308.el5`

sanchezzzhak (dismissed):

Tests/Parser/Client/fixtures/library.yml
```diff
@@ -528,3 +527,33 @@
...
+  user_agent: python-urllib3/1.26.9
+  client:
+    type: library
+    name: Python urllib
+    version: 1.26.9
+-
+  user_agent: msray-plus
+  client:
+    type: library
+    name: msray
```
Simbiat, line 546: It looks like this is https://www.msray.net/doc/ and it should be named `Msray-plus`

Tests/Parser/Client/fixtures/library.yml
```diff
@@ -528,3 +527,33 @@
...
+    version: ""
+-
+  user_agent: Slim Framework
+  client:
+    type: library
+    name: Slim
```
Simbiat, line 552: Would not "Slim Framework" be a more appropriate name, since this is their official name?

Tests/Parser/Client/fixtures/library.yml
```diff
@@ -528,3 +527,33 @@
...
+    version: ""
+-
+  user_agent: Fuzz Faster U Fool v1.5.0-dev
+  client:
+    type: library
+    name: Fuzz Faster U Fool
```
Simbiat, line 558: FFUF seems to be used way more often, and since official logo uses abbreviation, I wonder if the name should be FFUF, too

Tests/fixtures/bots.yml
```diff
@@ -5828,3 +5828,48 @@
...
+    name: Matomo
+    category: Service Agent
+    url: https://github.com/matomo-org/matomo
+    producer:
+      name: InnoCraft Ltd
+      url: https://matomo.org/
+-
+  user_agent: CustomUserAgent/1.0
+  bot:
+    name: Generic Bot
```
Simbiat, line 5843: Do we really need this? This is essentially an unknown bot, and the library already handles unrecognized bots fine.

Tests/fixtures/bots.yml
```diff
@@ -5828,3 +5828,48 @@
...
+    name: Prometheus
+    category: Service Agent
+    url: https://github.com/prometheus/prometheus
+    producer:
+      name: The Linux Foundation
+      url: https://www.cncf.io/
+-
+  user_agent: firefox
+  bot:
+    name: Generic Bot
```
Simbiat, line 5856: Do we really need this? This is essentially an unknown bot, and the library already handles unrecognized bots fine.

Tests/fixtures/bots.yml
```diff
@@ -5828,3 +5828,48 @@
...
+-
+  user_agent: Chrome
+  bot:
+    name: Generic Bot
```
Simbiat, line 5860: Do we really need this? This is essentially an unknown bot, and the library already handles unrecognized bots fine.

liviuconcioiu:
> While reviewing this, I've learnt that we have "Generic bot" entry. I wonder if we actually need it, since this is equivalent to "Unknown" in my mind.

It's like a partial detection. We don't know which bot or for some bots there's no point adding them or there's not enough info. That's why this generic detection is used. Without it bots containing `spider` and `Chrome` would be detected as real visitors, which is false.

sanchezzzhak (dismissed):

sanchezzzhak (dismissed):

regexes/bots.yml
sgiehl: Good question. I would say it shouldn't.

sanchezzzhak (approved):

PR #7529, opened 2023-12-08
Detect new brands: GTX, Maunfeld, N-one, SAILF and Detect devices for exist brands

sanchezzzhak:


liviuconcioiu:
> I don't even know how to fix PHPCS, need help.

Not sure, but probably has something to do with https://github.com/squizlabs/PHP_CodeSniffer/issues/3932

sgiehl:
I've fixed that for now.

sgiehl (approved):

PR #7562, opened 2024-01-23
Detect new brands: MeMobile, VOLIA and Detect device for exists brands

sanchezzzhak:


Parser/Device/AbstractDeviceParser.php
```diff
@@ -992,6 +992,7 @@ abstract class AbstractDeviceParser extends AbstractParser
         'M3'  => 'Mecer',
         'M3M' => 'M3 Mobile',
         '0M'  => 'Mecool',
+        'MEM' => 'Me Mobile',
```
Simbiat, line 995: It looks like official name is `MeMobile" without a space. At least the closet to official seems to be their page https://www.facebook.com/memobilepakistan/ (their website is not loading for me). Media seems to be mostly using it with space, but I've found some pages without it, too.

Simbiat: also youtube channel https://www.youtube.com/channel/UC-FA0BPxjtVhkfg0IknOawg

Simbiat: Huh, their official website did load for after like 5 minutes of loading (per page), and I finally got to http://www.memobile.pk/legal/ where they are not using space in the name. I would consider legal name to be the "correct" one.

sanchezzzhak: fix

sgiehl (approved):

PR #7566, opened 2024-01-26
More podcast updates, rename NPR, fix ms project bot detection.

mindreader:
### Description:

More things from the opawg 2 podcast useragent list. Major things:

Rename NPR One -> NPR
Youtube Music detection
Some Amazon Echo detection
Microsoft Project was being detected as a bot, there was a test that missed this because that test doesn't test for botness
Various mobile apps, libraries, and a couple bots

### Review

* [*] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [*] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [*] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [*] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [*] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)


mindreader:
Checks seemed to pass, there was one odd cancellation, not sure what's up with that.

Tests/fixtures/podcasting.yml
```diff
@@ -9069,3 +9069,335 @@
...
+    brand: 'Samsung'
+    model: 'Galaxy A50'
+  os_family: Android
+  browser_family: Unknown
+
+-
+  user_agent: 'SM-T970 (compatible; Tablet2.0) HandelsbladProduction, com.twipemobile.nrc 5.1.4 (511) / Android 33'
+  os:
+    name: Android
+    version: "33"
```
liviuconcioiu, line 9259: The version here is wrong and it refers to `Android API 33` -> `Android 13`. I wanted to fix this a while ago. If you want to fix it, is great! Another user agent to be added as test:

`MXPlayer/1.73.3/2001002183; Android/33`

Tests/fixtures/podcasting.yml
```diff
@@ -9069,3 +9069,335 @@
...
+    brand: Apple
+    model: ''
+  os_family: iOS
+  browser_family: Unknown
+
+-
+  user_agent: 'Amazon;Echo_Dot;27d4dfe427b34d57995b463e5d63198d;;tpapi;3.199.422'
+  os: [ ]
+  client:
+    type: mobile app
```
Simbiat, line 9169: Would it not be more like a library? https://github.com/alandtse/alexa_media_player

Tests/fixtures/podcasting.yml
```diff
@@ -9069,3 +9069,335 @@
...
+
+-
+  user_agent: 'Dalvik/2.1.0 (Linux; U; Android 13; SM-G991U Build/TP1A.220624.014) nprone_android/3.7.0/OSv:13'
+  os:
+    name: Android
+    version: "13"
+    platform: ""
+  client:
+    type: mobile app
+    name: NPR
```
Simbiat, line 9280: Since other items were renamed to NPR One, is this one different somehow?

mindreader: The original 'NPR One' designation came from opawg2. The change to NPR also came from them. They reference this article, and indeed I cannot find any app any longer called NPR One on the android app store.

https://www.npr.org/about-npr/1218775096/npr-unifies-mobile-app-experience-for-listeners-sunsets-npr-one-app

Tests/fixtures/podcasting.yml
mindreader: I put in a fix for this, I hope duplicating the android block is acceptable, I don't think trying to combine those top level regexes seems like a very good idea.

Tests/fixtures/podcasting.yml
Simbiat: Ok, so old apps were NPR One, and new one is NPR, makes sense.

Tests/fixtures/podcasting.yml
Simbiat: If I am reading this right https://developer.amazon.com/en-US/docs/alexa/avs-device-sdk/mediaplayer.html and https://developer.amazon.com/en-US/docs/alexa/device-apis/alexa-media-playback.htmlit does not look like there ever was an official media player in Alexa, it's just that Alexa provided API to to have a media control (which explains the `tpapi` portion of the user agent). but even if we consider that this is not a library for home assistant, we do have a `mediaplayer` type.

mindreader: Interesting. Okay I guess I'll change it to Alexa, and change the client type to mediaplayer.  I never noticed it was mobile app. And I'll look at that home assistant project to ensure I'm not changing some user agent that it uses. If it is distinct, then I'll try and disambiguate it as a library client type, with some mention of home assistant.

regexes/oss.yml
```diff
@@ -429,6 +429,46 @@
 - regex: '(?:(?:Orca-)?(?<!like |/|RadioPublic |Anghami )Android|Adr|AOSP)[ /]?(?:[a-z]+ )?(\d+[\.\d]*)'
   name: 'Android'
   version: '$1'
+  versions:
+    - regex: 'Android[ /]34(?!.)'
```
sanchezzzhak, line 433: create a separate rule above
```yml
- regex: 'Android[ /]\d{2}$'
  name: 'Android'
  versions:
    - regex: 'Android[ /]34'
      version: '14'
````
etc...
this will reduce the number of iterations in the loop.
if you look at the code, you will understand everything:

https://github.com/matomo-org/device-detector/blob/59fb555c303eac7cd8fe6e9d0a56347faa2a3ad5/Parser/OperatingSystem.php#L508-L534

mindreader: After reading your comment, I realized I could combine the sdk version blocks, and lead to better code overall. That was not easy, some of the versions it is almost down to looking to see what range the version is in, but I think I succeeded without writing any php.

Tests/fixtures/podcasting.yml
mindreader: Updated.

sanchezzzhak:
Fixed #7643

PR #7567, opened 2024-01-27
Fix engine detection on old Qwant iOS versions.

JBYoshi:
### Description:

Until just this week, all browsers on iOS were required to use WebKit. However, I saw that a test case for the Qwant browser on iOS was incorrectly being detected as using the Gecko engine. This PR updates the `QwantMobile` token detection to allow for different engines.

### Review


Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -2012,8 +2012,8 @@
     type: browser
     name: Qwant Mobile
     version: "1.3"
-    engine: Gecko
-    engine_version: ""
+    engine: WebKit
+    engine_version: "604.1.38"
     family: Firefox
```
JBYoshi, line 2017: The browser family is still being marked as Firefox here (since all the Qwant Mobile entries use the same browser and that browser is defined as part of the Firefox family). Should I change that? If so, what should I change it to?

sanchezzzhak: for iOS, everything that is a browser is always a Webkit engine, more precisely safari. Therefore, it does not matter which browser you put on your phone, it will be safari and have similar +- inconveniences for developers and the same display errors.

so the current changes for iOS are correct

Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -2786,8 +2786,8 @@
     type: browser
     name: Qwant Mobile
     version: "1.1"
-    engine: Gecko
-    engine_version: ""
+    engine: WebKit
```
sanchezzzhak, line 2789: engine: blink
family: Chrome

JBYoshi: For the family part, this is the same issue as I mentioned for iOS - the family is only specified in terms of the browser name, so the family detector can't distinguish between a Gecko-based Qwant Mobile and a Chrome-based Qwant Mobile. The best option I can think of would be to set the family to null for all Qwant browsers.

The engine mismatch seems to be affecting a lot of Blink-based browsers which are all currently being detected as WebKit 537.36. I was planning to do a separate PR to handle that but it'll be a large one.

sanchezzzhak: We solved this for other browsers by creating an additional rule.
something like this:
```yml
- regex: 'Chrome/.+QwantMobile/(\d+\.[.\d]+)'
  name: 'Qwant Mobile'
  version: '$1'
  engine:
    default: 'Blink'
```

sanchezzzhak (approved):

Tests/Parser/Client/fixtures/browser.yml
sanchezzzhak: We may need to add additional code in the future.
If the browser does not belong to the browser family,
check the presence of the words Firefox, Chrome in the user agent and compare them with the browser family
For Ios, leave browser_family: unknown
Also take care of Yandex browser_family: unknown (although it has been running on the chrome engine for long time)

this hypothesis requires detailed study.

Issue #7544, opened 2023-12-25
How about adding more device types?


sanchezzzhak:
I suggest adding sub device types (sub-types will not affect the current baseline result, but will give specifics about the type of device):

|  sub type  |  base type |
| :--- | :--- |
| post scanner  | `smartphone` or `peripheral` |
| post terminal  | `smartphone` or `table` or  `peripheral` |
| home cinema| tv |
| projector | tv | |
| laptop| desktop |
| nootebok | desktop |
| printer | peripheral |
| ebook reader | tablet |
| watch | watches | |
| headset | watches |
| glasses | watches |

for those who use the `device.type` will not change anything.

liviuconcioiu:
I'm not against.

EfekanPekel:
It would be nice.
The set-top boxes (STB) can also be added under the TV as a sub_type.

sgiehl:
As long as we only introduce sub types we can do that any time. But we should avoid changing the existing types. If that would be required we need to release a new major version for that.

Simbiat:
There are #7488, #7487, #7489 and potentially #7491 that would benefit from a major release. I wanted to address those in respectful PRs, but I got myself into going through brands (898 parsed) and I have also found at least 91 "issues" (wrong cases in names, entirely wrong brands, incomplete names, etc.), so I thought to finish those first. Then there is a new #7571 I've just raised, that we probably need to find a solution for.
What I'm trying to say is there are quite a few relatively significant changes, so I think it's better to not use subtypes, but to add new types instead, and then combine all these changes into a new major release. Not sure if you want to wait for me to finish all brands (875 left), let alone bots, though.

Why I think it's better to have new types? Well, at the least projectors are not TVs, they are not even close, even though you do get an image in the end, ebook readers are not necessarily tablets, there are those the size of a phone, too, and headsets and glasses are definitely not related to watches. In addition, it may be a good idea to think of a way to allow frictionless addition of new types of the devices, too. It would be more straightforward to maintain that in the long run compared to maintaining a list of sub types, where we may start debating whether a device is type A or type B (like Apple Vision: is it glasses or peripheral or "PC"?).

PR #7570, opened 2024-01-30
allow later Symfony versions

tacman:
### Description:

### Review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "matthiasmullie/scrapbook": "^1.4.7",
         "phpstan/phpstan": "^1.10.44",
         "mayflower/mo4-coding-standard": "^v9.0.0",
-        "symfony/yaml": "^5.1.7"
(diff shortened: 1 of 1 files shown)


sgiehl:
Some tests seem to be failing. Maybe newer version aren't compatible with our code 🤔

tacman:
Yes, it appears to fail in PHP 7.2, which is so old it doesn't even show up on https://www.php.net/supported-versions.php

How about releasing a version 7 of this library that drops support for anything EOL?  A bonus is that then the code could up updated to use some of the goodness that comes with PHP 8.1+, like typehints and property promotion.  Anyone still using PHP 7.2 could continue to use version 6.

Simbiat:
I guess that would be another reason for new major version added to the list in #7544

tacman:
Yes, I believe that would work, though I'm still learning how to properly set up the matrix for github actions.

sgiehl:
you can try changing the composer installs in the linked file above and remove the ` --ignore-platform-reqs`

PR #7572, opened 2024-02-01
Improves version detection for iOS, macOS and Iridium

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -820,6 +820,7 @@ public function parse(): ?array
                 || 0 === \strpos($version, '2021')
                 || 0 === \strpos($version, '2022')
                 || 0 === \strpos($version, '2023')
+                || 0 === \strpos($version, '2024')
```
sanchezzzhak, line 823: can we combine rules into a regular expression?

```php
if (\preg_match('/202[0-4]/', $version)){
     $name          = 'Iridium';
     $short         = 'I1';
}
```

sanchezzzhak: or 
```php
if (\preg_match('/^202[0-4]/', $version)){
     $name          = 'Iridium';
     $short         = 'I1';
}
```

sanchezzzhak (approved):

PR #7574, opened 2024-02-07
Adds detection for BrightSign operating system, improves detection for Android, iPadOS, iOS, macOS operating systems, adds detection for various bots, apps, libraries and browsers

liviuconcioiu:


Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -7937,3 +7937,213 @@
...
+    engine: Blink
+    engine_version: 103.0.5060.129
+    family: Chrome
+  headers:
+    http-x-requested-with: com.qmamu.browser
+-
+  user_agent: Mozilla/5.0 (Linux; Android 12) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/111.0.5563.58 Mobile DuckDuckGo/17 Safari/537.36
+  client:
+    type: browser
+    name: BF Browser
```
Simbiat, line 8132: We have a `BF Browser` registered with `com.bf.browser`, which now is called `BXE Browser`. They are not related, though, so I guess it would be better to also rename `com.bf.browser` while adding this one.
At the same time there are also two clones with `com.befaster.bfbrowser` and `com.befaster.bfbrowser`. They are definitively clones of each other, but I also think they are clones of this this new BF Browser, but slightly different versions. I mean `com.techlastudio.bfbrowser` description is largely the same as one of those clones and, and it does say "the latest version", which may mean that it's the latest clone. It may be worth it to add IDs from the older versions.

Tests/Parser/fixtures/oss.yml
```diff
@@ -4142,3 +4142,43 @@
...
+  os:
+    name: iOS
+    short_name: IOS
+    version: "15.0"
+    platform:
+    family: iOS
+-
+  user_agent: BrightSign/8.3.23 (XT1144) Mozilla/5.0 (X11; Linux aarch64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.3 Chrome/69.0.3497.128 Safari/537.36
+  os:
+    name: BrightSign
```
Simbiat, line 4156: Looks like official name is `BrightSignOS` https://www.brightsign.biz/ecosystem/brightsign-os/

Tests/fixtures/clienthints-app.yml
```diff
@@ -2233,3 +2233,231 @@
...
+-
+  user_agent: Mozilla/5.0 (Linux; Android 8.0.0; SM-G935F Build/R16NW; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/98.0.4758.101 Mobile Safari/537.36
+  headers:
+    http-x-requested-with: org.lilo.app
+  os:
+    name: Android
+    version: 8.0.0
+    platform: ""
+  client:
+    type: mobile app
```
Simbiat, line 2245: We have Lilo from `org.lilo.mobile.android2020` and it's a browser

Tests/fixtures/clienthints-app.yml
```diff
@@ -2233,3 +2233,231 @@
...
+    name: Lilo
```
Simbiat, line 2246: `org.lilo.browser` has `Lilo Browser` as its name on PlayStore

Tests/fixtures/clienthints-app.yml
```diff
@@ -2233,3 +2233,231 @@
...
+  user_agent: Mozilla/5.0 (Linux; Android 10; 5007G Build/QP1A.190711.020; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/88.0.4324.181 Mobile Safari/537.36
+  headers:
+    http-x-requested-with: com.nocardteam.nocardvpn
+  os:
+    name: Android
+    version: "10"
+    platform: ""
+  client:
+    type: mobile app
+    name: NoCard VPN
```
Simbiat, line 2390: Current name is `NoCard Lite` https://play.google.com/store/apps/details?id=com.nocardteam.nocardvpn.lite&hl=en_US

Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -7937,3 +7937,213 @@
...
+    engine: Blink
+    engine_version: 100.0.4896.88
+    family: Chrome
+  headers:
+    http-x-requested-with: com.yandex.browser.alpha
+-
+  user_agent: Mozilla/5.0 (Linux; Android 11; SM-S506DL Build/RP1A.200720.012; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/101.0.4951.61 Mobile Safari/537.36
+  client:
+    type: browser
+    name: Samsung Browser
```
Simbiat, line 8026: Should we not differentiate this one (`Samsung Browser Beta`), since this is beta version, which is clear both from ID and name in PlayStore? I mean, since Samsung released this as a completely separate entity instead of using "beta program" or whatever it's called

Tests/fixtures/clienthints-app.yml
Simbiat: oh, I did notice that the `lite` was added to the ID (even though I googled `"com.nocardteam.nocardvpn"`). But, yes, `NoCard VPN Lite` sounds good.
Would it not make sense to split VPNs into separate category? Although that would also mean that we need to split podcast apps, and probably some other types, too.

Tests/Parser/Client/fixtures/browser.yml
liviuconcioiu: I've had the same idea a few years ago, but hasn't been approved. See https://github.com/matomo-org/device-detector/pull/6073#issuecomment-544256158

Tests/fixtures/clienthints-app.yml
```diff
@@ -2233,3 +2233,231 @@
...
+-
+  user_agent: Mozilla/5.0 (Linux; Android 10; SM-G965U Build/QP1A.190711.020; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/99.0.4844.88 Mobile Safari/537.36 DaumApps/6.9.27 DaumDevice/mobile
+  headers:
+    http-x-requested-with: net.daum.android.daum
+  os:
+    name: Android
+    version: "10"
+    platform: ""
+  client:
+    type: mobile app
```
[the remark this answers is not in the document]
liviuconcioiu: It's more like an all-in-one app, like WeChat.

regexes/client/hints/apps.yml
```diff
@@ -125,3 +135,6 @@
 'com.surfshark.vpnclient.android': 'Surfshark'
 'com.omshyapps.vpn': 'Omshy VPN'
 'com.kuto.vpn': 'KUTO VPN'
+'com.ezt.vpn': 'EZVPN'
+'com.nocardteam.nocardvpn': 'NoCard VPN Lite'
```
sanchezzzhak, line 139: `NoCard VPN`

sanchezzzhak (approved):

PR #7598, opened 2024-02-17
Adds detection for Telkom, UD brands, improves detection for Hyundai, Nordmende, O2, Orange and unknown brands, adds detection for QbHttp library and TV Bro browser

liviuconcioiu:


Tests/fixtures/clienthints-app.yml
```diff
@@ -2461,3 +2461,21 @@
...
+  user_agent: Mozilla/5.0 (Linux; Android 11.0; Orange TV Box Build/QTG1.200615.001; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/102.0.5005.22 Mobile Safari/537.36
+  headers:
+    http-x-requested-with: com.droidlogic.xlauncher
+  os:
+    name: Android
+    version: "11.0"
+    platform: ""
+  client:
+    type: mobile app
+    name: Droidlogic Launcher
```
Simbiat, line 2474: Are we sure about the name? There are not many mentions of the `com.droidlogic.xlauncher`, but I found this https://4pda.to/forum/index.php?showtopic=1052868&st=560#entry117455041 post and you can see on screenshot, that this ID is called `MediaBox launcher` but even then I am not sure it's correct one, since this seems to be for a different box

liviuconcioiu: There are already a few apps like this:

```
'com.tct.launcher': 'Joy Launcher'
'com.transsion.XOSLauncher': 'XOS Launcher'
'com.transsion.hilauncher': 'HiOS Launcher'
'com.transsion.itel.launcher': 'itel Launcher'
```

Simbiat: Of the four you mentioned 3 exist on Google Play, and `com.tct.launcher` exists on APKMirror and Uptodown at least, so I have no concerns about them. `com.droidlogic.xlauncher` does not seem to exist anywhere aside 2 mentions on 4PDA and 1 mention on a Chinese forum, which is also about removing of this thing (https://www.znds.com/tv-1231796-1-1.html). That's why I am not sure about the name, especially since on 4PDA it is named differently.
On the other hand googling for `"com.droidlogic.xlauncher"` (with quotes) in images still suggests https://play.google.com/store/apps/details?id=com.model.x.launcher which is a launcher, but not for TV and https://dimitrology.com/sdm_downloads/mediabox-launcher/ which does say `MediaBox Launcher`, same as 4PDA and it does have an APK even, but inside of it is `com.amlogic.mediaboxlauncher`

liviuconcioiu: I think they produce TV boxes for other brands - https://droidlogic.tv/

Simbiat: I am not questioning Droidlogic, I am questioning the name of the app, since little evidence that is there suggests a different name for it.

liviuconcioiu: I've changed it to `X Launcher`.

sanchezzzhak (approved):

PR #7608, opened 2024-03-01
Improves detection for wearable devices

liviuconcioiu:


DeviceDetector.php
```diff
@@ -793,6 +793,18 @@ protected function hasAndroidMobileFragment(): bool
...
     }
 
+    /**
+     * Returns if the parsed UA contains the 'Android; Mobile VR;' fragment
+     *
+     * @return bool
+     */
+    protected function hasAndroidVRFragment(): bool
+    {
+        $regex = 'Android( [\.0-9]+)?; Mobile VR;';
```
sanchezzzhak, line 803: mb merge regex one rule?
```php
$regex = 'Android( [\.0-9]+)?; Mobile VR;| VR ';
```
and 

```php
if (null === $this->device && ($this->hasAndroidVRFragment()) {
```

DeviceDetector.php
```diff
@@ -774,7 +774,7 @@ protected function getOsAttribute(string $attr): string
      *
      * @return bool
      */
-    protected function hasAndroidTableFragment(): bool
+    protected function hasAndroidTabletFragment(): bool
```
sanchezzzhak, line 777: To avoid mistakes in projects, who could use this method.
I suggest safe the old name `hasAndroidTableFragment`

sanchezzzhak (approved):

PR #7622, opened 2024-03-09
Adds detection for various browsers and improves version detection for Ubuntu

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -92,9 +93,11 @@ class Browser extends AbstractClientParser
         'A5' => 'Amerigo',
         'AO' => 'Amigo',
         'AN' => 'Android Browser',
+        '3A' => 'AOL',
```
Simbiat, line 96: According to https://browserwars.fandom.com/wiki/AOL_Explorer and https://en.wikipedia.org/wiki/AOL_Explorer (which clearly indicates Trident engine) and some screenshots I was able to find, it looks like this should be `AOL Explorer` or, at least `AOL Browser`

Tests/fixtures/console.yml
```diff
@@ -564,3 +564,33 @@
...
+  client:
+    type: browser
+    name: Aplix
+    version: "1.03"
+    engine: ""
+    engine_version: ""
+  device:
+    type: console
+    brand: Sanyo
+    model: ""
```
sanchezzzhak, line 594: most likely it's a game console `3DO TRY`

Parser/Client/Browser.php
```diff
@@ -92,9 +93,11 @@ class Browser extends AbstractClientParser
         'A5' => 'Amerigo',
         'AO' => 'Amigo',
         'AN' => 'Android Browser',
+        '3A' => 'AOL Browser',
```
sanchezzzhak, line 96: AOL Explorer

image

regexes/device/consoles.yml
```diff
@@ -35,10 +35,19 @@ OUYA:
...
   model: 'OUYA'
 
+Sanyo:
+  regex: '.*_SANYO_.*'
+  device: 'console'
+  model: ''
+
 Sega:
-  regex: 'Dreamcast'
+  regex: 'Dreamcast|.*_SEGASATURN_.*'
```
sanchezzzhak, line 44:
```
regex: 'Dreamcast|Aplix_SEGASATURN'
```
greedy modifiers are best avoided if possible (there is only 1 browser under sega saturn, and then it was accessed via a cartridge/disk)

sanchezzzhak (approved):

PR #7630, opened 2024-03-18
Adds detection for ASPLinux, SerenityOS, improves detection for Fire OS, improves version detection for Fire OS, Lineage OS and tvOS

liviuconcioiu:


liviuconcioiu:
Lineage OS versions from each version branch `https://github.com/LineageOS/android/blob/lineage-21.0/default.xml`

Parser/OperatingSystem.php
```diff
@@ -305,6 +307,45 @@ public function parse(): ?array
...
+            '6.0.1' => '13.0',
+            '6.0'   => '13.0',
+            '5.1.1' => '12.1',
+            '5.0.2' => '12.0',
+            '5.0'   => '12.0',
+            '4.4.4' => '11.0',
+            '4.3'   => '10.2',
+            '4.2.2' => '10.1',
+            '4.0.4' => '9.1.0',
+        ];
```
sanchezzzhak, line 347: I suggest creating 2 private property,  since the function body has already become large

```
private $fireOsVersionMapping = [];

private $lineageOsVersionMapping = [];
```

sanchezzzhak (approved):

PR #7640, opened 2024-03-24
Adds detection for ElectroBSD and improves detection for Amazon Linux, Android, Chrome OS, iPadOS, iOS, Lineage OS, macOS

liviuconcioiu:


sanchezzzhak (dismissed):

Parser/OperatingSystem.php
```diff
@@ -280,7 +284,7 @@ class OperatingSystem extends AbstractParser
      * @var array
      */
     private $lineageOsVersionMapping = [
-        '14'    => '21.0',
+        '14'    => '21',
```
sgiehl, line 287: Is there a specific reason for removing the `.0`?

liviuconcioiu: > Is there a specific reason for removing the `.0`?

https://lineageos.org/Changelog-28/
_As Android has moved onto the quarterly maintenance release model, this release will be “LineageOS 21”, not 21.0 or 21.1 - though worry not - we are based on the latest and greatest Android 14 version, QPR1._

regexes/oss.yml
```diff
@@ -1233,15 +1248,15 @@
...
+- regex: 'iOS[ /]Version ([\d\.]+)'
   name: 'iOS'
   version: '$1'
 
 - regex: 'Sonos/.+\(ICRU_'
   name: 'iOS'
   version: ''
 
-- regex: 'CaptiveNetworkSupport|AirPlay'
+- regex: 'CaptiveNetworkSupport|AirPlay|.*[ \.\-/]iOS'
```
sanchezzzhak, line 1259: `.*[ ./-]iOS`

EA:
Redundant character escape '\.' in RegExp 
 Inspection info: Reports redundant character escape sequences that can be replaced with unescaped characters preserving the meaning. Many escape sequences that are necessary outside of a character class are redundant inside square brackets [] of a character class.
Although unescaped opening curly braces { outside of character classes are allowed in some dialects (JavaScript, Python, and so on), it can cause confusion and make the pattern less portable, because there are dialects that require escaping curly braces as characters. For this reason the inspection does not report escaped opening curly braces.
Example:
  \-\;[\.]
After the quick-fix is applied:
  -;[.]
  
  
  This is not a mistake, we will put everything in order in the future

sanchezzzhak (approved):

PR #7636, opened 2024-03-23
Clean test files

liviuconcioiu:
Fixes #7633

PR #7652, opened 2024-04-06
Cleanup tests

liviuconcioiu:
Remove double quotes from version if isn't a float and replaces single quotes with double quotes, especially in podcasting.yml.


sgiehl:
@liviuconcioiu the double quotes around the version numbers had been added on purpose at some point. If I remember correctly some of the forks had problems with parsing the yml files otherwise, as the built in yml parser couldn't parse it without.

liviuconcioiu:
> @liviuconcioiu the double quotes around the version numbers had been added on purpose at some point. If I remember correctly some of the forks had problems with parsing the yml files otherwise, as the built in yml parser couldn't parse it without.

So what do you suggest to do in this case? Should I add double quotes to all versions? Right now there's a mix and I want to fix this.

sgiehl:
> Also see @sanchezzzhak comment here https://github.com/matomo-org/device-detector/pull/7636#issuecomment-2020302013

the comment applies to the yaml parser we are using in php, as it automatically tries to determine the type.
I would personally say we should add surrounding `"` to all values that are a string, but might contain numerous values.

sanchezzzhak:
I use a very strict YAML parser in nodejs
for example a format fixture
```yml
-
  user_agent: Android 12.0
  os: 
    name: Android
    version: 12.0 
```
nodejs parse result:
```js
   {
        user_agent: "Android 12.0",
        os: {
            name: "Android",
            version: 12  // Number type
       }
   }
```
php parse result:
```php
   [
        "user_agent"  =>  "Android 12.0",
        "os" => [
            "name": "Android",
            "version": "12.0"
       ]
   ]
```
the current fixtures are all fixed and at the moment I am satisfied, as they do not give errors in the tests
https://github.com/sanchezzzhak/node-device-detector/actions/runs/8597281681/

total:
you don't need to redo everything now, just make sure that if there is a float/int value in the fixtures, it should be wrapped in quotation marks.
if there are square brackets in the string, this value should also be wrapped in a string.
if you use the script `php misc/test.php "useragent"` there will never be any problems, he creates the fixtures perfectly.

liviuconcioiu:
> the script `php misc/test.php "useragent"` there will never be any problems, he creates the fixtures perfectly.

This PR will do exactly this, so all tests are similar in formatting.

sanchezzzhak (approved):

PR #7663, opened 2024-04-20
Replaces $1 from generated readme with actual clients

liviuconcioiu:


Parser/Client/AbstractClientParser.php
```diff
@@ -82,6 +82,31 @@ public static function getAvailableClients(): array
...
             }
 
+            if ('Microsoft Office $1' === $regex['name']) {
+                $clients = [
+                    'Microsoft Office Access', 'Microsoft Office Excel', 'Microsoft Office OneDrive for Business',
+                    'Microsoft Office OneNote', 'Microsoft Office PowerPoint', 'Microsoft Office Project',
+                    'Microsoft Office Publisher', 'Microsoft Office Visio', 'Microsoft Office Word',
+                ];
+
+                $regex['name'] = \implode(', ', $clients);
```
sanchezzzhak, line 92: I think we should do it like this:
```php
foreach($clients as $client) {
   $names[] = $client;
}
continue;
```
this will not cause confusion when I want to check if there is such a client in the array
```php
in_array('Microsoft Office Publisher', MobileAppParser::getAvailableClients())
```

sanchezzzhak (dismissed):

Parser/Client/AbstractClientParser.php
```diff
@@ -82,6 +82,43 @@ public static function getAvailableClients(): array
...
+
+                foreach ($clients as $client) {
+                    $names[] = $client;
+                }
+
+                continue;
+            }
+
             $names[] = $regex['name'];
         }
```
sanchezzzhak, line 123: I suggest just combining the array instead of 3 cycles

```php
    public static function getAvailableClients(): array
    {
        $instance = new static(); // @phpstan-ignore-line
        $regexes  = $instance->getRegexes();
        $names    = [];

        foreach ($regexes as $regex) {
            if (\strpos($regex['name'], '$1') !== false) {
                continue;
            }

            $names[] = $regex['name'];
        }

        if (static::class === MobileApp::class) {
            $names = \array_merge($names, [
                // Microsoft Office $1
                'Microsoft Office Access', 'Microsoft Office Excel', 'Microsoft Office OneDrive for Business',
                'Microsoft Office OneNote', 'Microsoft Office PowerPoint', 'Microsoft Office Project',
                'Microsoft Office Publisher', 'Microsoft Office Visio', 'Microsoft Office Word',
                // Podkicker$1
                'Podkicker', 'Podkicker Pro', 'Podkicker Classic',
                // radio.$1
                'radio.at', 'radio.de', 'radio.dk', 'radio.es', 'radio.fr',
                'radio.it',  'radio.pl', 'radio.pt', 'radio.se',  'radio.net',
            ]);
        }

        \natcasesort($names);

        return \array_unique($names);
    }   
```

sanchezzzhak (approved):

PR #7665, opened 2024-04-21
Improves detection for mobile devices

liviuconcioiu:


regexes/device/mobiles.yml
```diff
@@ -21282,8 +21282,12 @@ Samsung:
       model: 'Galaxy S23 Ultra'
     - regex: '(?:SAMSUNG-)?(?:SM-S711)'
       model: 'Galaxy S23 FE'
+    - regex: '(?:SAMSUNG-)?SM-S921(?:U1|[0BNQUW])'
+      model: 'Galaxy S24'
     - regex: '(?:SAMSUNG-)?(?:SM-S926)'
       model: 'Galaxy S24+'
+    - regex: '(?:SAMSUNG-)?SM-S928(?:U1|[0BNQUW])'
```
sanchezzzhak, line 21289: `(?:SAMSUNG-)?SM-S928`

there is no need to specify all the regions, there are actually more of them and even Google does not know them all  (only in case of collisions)
https://storage.googleapis.com/play_public/supported_devices.html

of-top:
> to create reports, I use this tool, which finds non-added user agents as tests, as well as if we have not identified the device.
```
git clone https://github.com/sanchezzzhak/node-device-detector
npm i
cd tests
; all help params node tools/report-test-files-ua.js -h 
node tools/report-test-files-ua.js /home/user/Downloads/xab -p strign -dt 1 -dc 1 -o string > output/xab.log
```
> (This allows you to find a large number of new devices that need to be processed. There are really a lot of them)

> I can make these reports as tasks, but I'm afraid there won't be enough space for public posting
1 day of logs is about 24GB of data

sanchezzzhak (changes requested):

sanchezzzhak (approved):

PR #7666, opened 2024-04-22
Detect new brands: MAG, S-Color and Detect devices for exist brands

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -4623,6 +4628,8 @@ Vivo:
       model: 'X9s'
     - regex: 'V2241A'
       model: 'X90'
+    - regex: 'V2219'
```
liviuconcioiu, line 4631:
```suggestion
    - regex: 'V2219(?:[);/ ]|$)'
```

liviuconcioiu (changes requested):

liviuconcioiu (approved):

PR #7671, opened 2024-04-27
Detect new brands: Cidea, Cuiud, Daria,  HiBy, mipo and Detect devices for exist brands

sanchezzzhak:


regexes/device/mobiles.yml
```diff
@@ -34127,6 +34190,11 @@ Datalogic:
   device: 'smartphone'
   model: 'Memor $1'
 
+# mipo (mipo.com.tr)
+mipo:
+  regex: 'mipo[ _]([^/;)]+)(?: Build|[);])'
+  model: '$1'
```
liviuconcioiu, line 34196:
```suggestion
  device: 'smartphone'
  model: '$1'
```

liviuconcioiu (changes requested):

Parser/Device/AbstractDeviceParser.php
```diff
@@ -329,6 +329,7 @@ abstract class AbstractDeviceParser extends AbstractParser
         'C3'  => 'China Mobile',
         'U9'  => 'China Telecom',
         'CI'  => 'Chico Mobile',
+        'CIA' => 'Cidea',
```
Simbiat, line 332: Based on photos I've found (for example https://jiji.ng/surulere/tablets/new-c-idea-cm7000-plus-512-gb-rsYzAEHJMsdZor3u7ZfnlZks.html) it should be `C Idea`

liviuconcioiu (approved):

PR #7674, opened 2024-04-29
Improves detection for masscan

### LINKED ISSUE #7673: Duplicate: 'masscan'
https://github.com/matomo-org/device-detector/blob/master/regexes/bots.yml#L964
https://github.com/matomo-org/device-detector/blob/master/regexes/client/libraries.yml#L610

liviuconcioiu:
Closes #7673


regexes/bots.yml
```diff
@@ -961,7 +961,7 @@
     name: 'BIZON, OOO'
     url: 'https://bi.zone/'
 
-- regex: 'masscan'
+- regex: '(?:ivre-)?masscan'
```
sgiehl, line 964: Can't we simply do the check without the additional optional check in the front? It doesn't have any value, or does it?

liviuconcioiu: It doesn't. I've added `.*`, otherwise, the built regex won't detect it.

sanchezzzhak (dismissed):

sgiehl (approved):

PR #7687, opened 2024-05-17
Adds detection for NOOK Browser

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -403,6 +403,7 @@ class Browser extends AbstractClientParser
         'MY' => 'Mypal',
         'MR' => 'Monument Browser',
         'MW' => 'MAUI WAP Browser',
+        'ZL' => 'MultiZilla',
```
Simbiat, line 406: Based on what I've found (http://www.os2voice.org/VNL/past_issues/VNL1202H/vnewsf5.htm and https://web.archive.org/web/20051231025133/http://multizilla.mozdev.org/) this is an extension, rather than a browser.

sanchezzzhak (approved):

PR #7688, opened 2024-05-17

liviuconcioiu:


Tests/Parser/Client/fixtures/library.yml
```diff
@@ -659,3 +659,15 @@
     type: library
     name: Boto3
     version: 1.17.18
+-
+  user_agent: Custom-AsyncHttpClient
```
Simbiat, line 663: Is this (and `Custom HTTP Client`) really a library? I am unable to find anything specific, feels more like a generic name, which should imply "unknown" user agent.

sanchezzzhak (approved):

PR #7715, opened 2024-06-20
Adds detection for Awesomium and Prism

liviuconcioiu:


Tests/Parser/Client/fixtures/mobile_app.yml
```diff
@@ -2069,3 +2069,15 @@
...
+-
+  user_agent: Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.6; en-US; rv:1.9.2.3) Gecko/20100402 Prism/1.0b4
+  client:
+    type: mobile app
+    name: Prism
+    version: "1.0"
+-
+  user_agent: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1003.1 Safari/535.19 Awesomium/1.7.4.2
+  client:
+    type: mobile app
```
Simbiat, line 2081: Not sure it's correct to treat this as mobile app, since this was a web-browser engine (even used it at some point). Or more like a library, that implemented browser-features. I am not aware of any full-fledged browsers, that used this, but there were quite a few apps, that used it for integrated browser windows, so maybe this should be treated as a browser by itself? Or maybe both as a browser and engine.

Tests/Parser/Client/fixtures/mobile_app.yml
```diff
@@ -2069,3 +2069,15 @@
     type: mobile app
     name: Zalo
     version: ""
+-
+  user_agent: Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.6; en-US; rv:1.9.2.3) Gecko/20100402 Prism/1.0b4
+  client:
+    type: mobile app
```
Simbiat, line 2075: From description of the product, it looks like it was more like a library allowing to integrate Gecko engine. Probably would make more sense as browser with Gecko engine.

Tests/Parser/Client/fixtures/mobile_app.yml
liviuconcioiu: That's what I thaught too. I had doubt about adding it as app. As far as I can see, it was also used for automation tools.

liviuconcioiu:
Closing in favor of #7713.

PR #7726, opened 2024-07-02
Adds detection for iNet Browser, Motorola Internet Browser, UPhone Browser, VMS Mosaic and improves detection for Dorado, NetFront, Nokia Browser

liviuconcioiu:


sanchezzzhak (approved):

sanchezzzhak:
It is better to add ancient browsers to the end of the list as much as possible, since the purity of such visits is minimal or zero.

liviuconcioiu:
> It is better to add ancient browsers to the end of the list as much as possible, since the purity of such visits is minimal or zero.

I've had an idea a few weeks ago to move well known browsers at the top, and see if there's some improvement gain. I'll try and measure if is worth it.

PR #7727, opened 2024-07-02
Make tests compatible with newer phpunit versions

williamdes:
### Description:

- Non breaking changes for newer phpunit versions (running on PHPUnit 11.1.3 at Debian)
- Fixing the parameter names for newer PHP versions (Error: Unknown named parameter $check)
- Upgrade GitHub actions
- Allow newer PHPunit versions up to 12
- Fix deprecations in tests

### Review

* [x] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [x] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [x] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [x] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [x] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [x] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [x] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [x] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [x] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [x] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [x] Existing documentation updated if needed


williamdes:
> Variable "user_agent" is not in valid camel caps format

Can someone give me the best way to fix this. Not by changing the variable name for sure 😄

sgiehl:
Why should the variable name need to contain a `_`? What is the problem with `useragent`? This doesn't comply with our coding style.

williamdes:
> Why should the variable name need to contain a `_`? What is the problem with `useragent`? This doesn't comply with our coding style.

Nope, because as you can see `getTypeMethodFixtures` is directly fed from `\Spyc::YAMLLoad($fixturePath);` 
So the variable names should match the yaml data key names

One solution is that I can put an array map to re-format the key name

williamdes:
Hello @sgiehl 

What is the best path for me to get this one merged ?

Tests/DeviceDetectorTest.php
```diff
@@ -638,16 +638,16 @@ public function getTypeMethodFixtures(): array
     /**
      * @dataProvider getTypeMethodFixtures
      */
-    public function testTypeMethods(string $ua, array $checkTypes): void
+    public function testTypeMethods(string $user_agent, array $check): void
```
sanchezzzhak, line 641:
```
public function testTypeMethods(string $userAgent, array $deviceTypes): void
```
or 
```
public function testTypeMethods(string $useragent, array $deviceTypes): void
```
`$check` - just $check as if we are expecting a bool value

williamdes: clearly nobody understands that it must match the data in the yaml files or it is a PHP incompatibility
So I will add an array map to rename the keys 😁

sanchezzzhak: We have PHPCS rules syntax in the project, they must be followed even in tests. We do not intend to change or weaken the policy for the tests, as well as the keys in the tests (since the fixture files are used in other projects for other programming languages).

The main developers are used to the current PHPCS rules, I even got used to it myself after ~7 years of adding tests.

williamdes: Yup I totally understand and would have said the same

And we can not change the data as others may use it and it would be a breaking change

Let's go to adapt the data to the policy on the fly

williamdes: I pushed https://github.com/matomo-org/device-detector/pull/7727/commits/b45667a7376cdf0789a1b90cf454137f4bf9ffc1 but I still have some work to do

williamdes:
Weird CI failures, news commits broke something or is it my diff ?

liviuconcioiu:
Not sure, but probably the PHPUnit workflow should be updated too so it can use `composer update`? Now it tries to run PHPUnit 11 on PHP 7.2.

I installed `PHP 7.2.5` (this is the lowest version required by `symfony/yaml`) and the tests pass.

.gitignore
```diff
@@ -5,3 +5,4 @@ composer.phar
 .buildpath
 .project
 .settings
+.phpunit.result.cache
```
williamdes, line 8:
```suggestion
```

composer.json
```diff
@@ -27,7 +27,7 @@
         "mustangostang/spyc": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^8.5.8",
+        "phpunit/phpunit": "^8.5.8 || ^9 || ^10 || ^11",
```
williamdes, line 30:
```suggestion
```

williamdes:
I dropped the last commit, it should help

liviuconcioiu:
@williamdes can you fix the conflicts? And also you should also add `static` to `getFixturesDeviceTypeFromClientHints`. It was added after your PR.

williamdes:
I did some ajustements as I am quite sure the ignore should not have been used for all versions and this way.

sanchezzzhak:
Run local php vendor/bin/phpcbf for fix phpcs

williamdes:
> Run local php vendor/bin/phpcbf for fix phpcs

Unrelated, see CI log

> PHP Parse error:  syntax error, unexpected '|', expecting variable (T_VARIABLE) in /home/runner/work/device-detector/device-detector/vendor/phpunit/phpunit/src/Framework/Assert/Functions.php on line 83

Probably some issue with the php version installed

liviuconcioiu:
@williamdes is it possible to update the workflow so it does look like this?

ok

Right now it looks like this:

notok

Those are missing (pending but aren't running):

PHPUnit (windows-latest, 8.4)
PHPUnit (macos-latest, 8.4)

williamdes:
> @williamdes is it possible to update the workflow so it does look like this?

Not really, can it be merged as is @sgiehl ?
I have one more PR for phpunit 12 afterwards

composer.json
```diff
@@ -27,7 +27,7 @@
         "mustangostang/spyc": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^8.5.8",
+        "phpunit/phpunit": "^8.5.2 || ^9 || ^10 || ^11",
```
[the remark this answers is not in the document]
williamdes: Here you go, I pushed it !

liviuconcioiu:
1. You should run `./vendor/bin/phpcbf` to fix the phpcs errors - https://github.com/matomo-org/device-detector/actions/runs/20045151285/job/57507861303?pr=7727

2. You can add to composer, so we get rid of `scrapbook` deprecations
```JSON
        "psr/cache": "^1.0.1 || ^2.0 || ^3.0",
        "psr/simple-cache": "^1.0.1 || ^2.0 || ^3.0",
```

```
Lock file operations: 0 installs, 3 updates, 0 removals
  - Upgrading matthiasmullie/scrapbook (1.4.9 => 1.5.4)
  - Upgrading psr/cache (1.0.1 => 3.0.0)
  - Upgrading psr/simple-cache (1.0.1 => 3.0.0)
```

3. On PHP 8.5 there are some deprecations - https://github.com/matomo-org/device-detector/actions/runs/20045151287/job/57507861394?pr=7727
```
3) E:\GitHub\device-detector\Tests\Parser\Client\BrowserTest.php:118
Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect

4) E:\GitHub\device-detector\Tests\Parser\Client\BrowserTest.php:122
Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect

5) E:\GitHub\device-detector\Tests\Parser\Client\BrowserTest.php:145
Method ReflectionMethod::setAccessible() is deprecated since 8.5, as it has no effect
```

Probably is a good idea to add this

```PHP

if (PHP_VERSION_ID < 80500) {
    $browserProperty->setAccessible(true);
}

```

sanchezzzhak (approved):

PR #7741, opened 2024-07-14
Adds detection for Spark, Total Browser, improves detection for Belva Browser, Phoenix and rename MIUI Browser to Mi Browser

liviuconcioiu:
- Spark - https://fastcorporate.com/
- Total Browser - https://www.totalbrowser.com/
- Mi Browser - https://trust.mi.com/docs/miui-privacy-white-paper-global/4/7


Parser/Client/Browser.php
```diff
@@ -570,6 +570,7 @@ class Browser extends AbstractClientParser
         'LE' => 'Smart Lenovo Browser',
         'OZ' => 'Smooz',
         'SN' => 'Snowshoe',
+        'K5' => 'Spark',
```
Simbiat, line 573: Googling seems to suggest that there have been only "Baidu Spark" browser (which we do have already). Is this the one or is it something different?

liviuconcioiu: Is another browser, made by https://fastcorporate.com/ - https://www.advanceduninstaller.com/Spark-e1f42f2b16bcd71251a972fdf9b3d716-application.htm

Parser/Client/Browser.php
```diff
@@ -393,7 +393,7 @@ class Browser extends AbstractClientParser
         'M3' => 'Midori Lite',
         'M6' => 'MixerBox AI',
         'MO' => 'Mobicip',
-        'MU' => 'MIUI Browser',
+        'MU' => 'Mi Browser',
```
sanchezzzhak, line 396: @sgiehl Do you agree to change the name?

sanchezzzhak (approved):

PR #7753, opened 2024-08-01
Cleanup tests #2

liviuconcioiu:
This is a follow up to #7652.

`oss.yml` - was missing from the script (#7752).
`clienthints.yml` - I forgot to clean this file in the initial PR.

The other commits are cleanups for commits merged after the initial PR.

### DIFF
--- a/Tests/Parser/Client/fixtures/browser.yml
+++ b/Tests/Parser/Client/fixtures/browser.yml
@@ -5188,9 +5188,9 @@
   client:
     type: browser
     name: CCleaner
-    version: "99.0.4844.51"
+    version: 99.0.4844.51
     engine: Blink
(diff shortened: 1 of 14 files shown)


sanchezzzhak (approved):

Tests/Parser/Client/fixtures/browser.yml
```diff
@@ -5188,9 +5188,9 @@
   client:
     type: browser
     name: CCleaner
-    version: "99.0.4844.51"
+    version: 99.0.4844.51
```
sgiehl, line 5191: This might actually cause problems for some of the forks to other languages. If I remember correctly some yaml parsers try to handle that as number when the `"` are left away.

PR #7756, opened 2024-08-03
Improves detection for Android and various libraries and adds detection for PhantomJS, Obisidian and Glory Star brand

liviuconcioiu:
I had to move libraries before browsers, otherwise it wouldn't detect `Cypress` and `Android License Verification Library`, and I fixed some regexes.

I also moved `Cypress` to libraries, since is a browser automation tool.


Tests/Parser/Client/fixtures/library.yml
```diff
@@ -689,3 +689,21 @@
...
+-
+  user_agent: Android.LVLDM
+  client:
+    type: library
+    name: Android License Verification Library
+    version: ""
+-
+  user_agent: Mozilla/5.0 (Unknown; Linux x86_64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.1.1 Safari/538.1
+  client:
+    type: library
```
Simbiat, line 701: should not this be a browser? project considered itself one, at least https://github.com/ariya/phantomjs

liviuconcioiu: It should, but I also saw that `SlimerJS` is also a library, instead of browser.

liviuconcioiu:
CLosing this one in favor of #7765, #7766, #7767 and #7768 for easier review.

PR #7766, opened 2024-08-08
Adds detection for PhantomJS, improves detection for Dezor and SlimerJS

liviuconcioiu:


sgiehl:
@liviuconcioiu I actually wouldn't count those tools as real browsers. They can be used to control / script a browser, but it is clearly not used as a typical browser.

liviuconcioiu:
Closing this in favor of #7797 and #7766.

PR #7770, opened 2024-08-09
Adds detection for myPhone (PL) and rename MyPhone to MyPhone (PH)

liviuconcioiu:
`myPhone` is different than `MyPhone`.


regexes/device/mobiles.yml
```diff
@@ -17829,7 +17829,15 @@ Rombica:
     - regex: '(Infinity K8|Smart Box (?:Q1|v00[3-9]|Quad|Ultra HD v002|4K V001|[ABCDXYZ]1|[AGDFQL]2|[AFG]3|H4)|TV (?:Ultima|Fly)|Smart Stick Pro|TV Mercury|TV Air)(?:[);/ ]|$)'
       model: '$1'
 
-# myphone (myphone.com.ph)
+# myPhone (https://www.myphone.pl/)
```
sanchezzzhak, line 17832: this is the same brand as below `MyPhone`

Simbiat: I do not think so. at least https://www.myphone.com.ph/ and https://www.myphone.pl/ have very different branding and different devices.
Due to name collision, though, I'd recommend to either rename old brand to `MySolid` (https://www.solidgroup.com.ph/about/business/my-solid-technologies-devices-corporation), or rename the new one to `TelForceOne` (https://www.myphone.pl/about/). Or add country to both: old one `MyPhone Philippines`, new one to `myPhone Poland`

sanchezzzhak: https://www.gsmchoice.com/ru/catalogue/myphone/models/40/

MyPhone - It also contains Polish models

It is best to create a separate brand and transfer models 
`MyPhone (PH)`:

myWX1 Plus
myXI1
myX12
my(?:XI[3]|NX1|X[23589]|A1[12356]|29S|807
myT[356] DTV
myA1
myA(10|[789]) DTV
myG1
myA10
myXI1(?:[ _](?:Plus|Pro))?

we cannot leave the brand in a different case, mysql selects data case-insensitive.

Simbiat: `mysql selects data case-insensitive` that depends on collation of the column.
but, IMHO, it would be better to either have `MyPhone (PH)` and `myPhone (PL)` then, or if we consider it fine to have country only for one of them - rename the newly added `myPhone` to `myPhone (PL)`. Not sure it's worth moving the old one, since way more devices.

liviuconcioiu: I've noticed on the screen logo, it says `MyPhone Pilipinas`.

https://www.myphone.com.ph/wp-content/uploads/2020/11/myWX1-Lite-blue.png

Simbiat: that is also an option, I guess. Or maybe use the pipe symbol? Even [Wiki](https://en.wikipedia.org/wiki/MyPhone) mentions that it's stylized as `my|phone`

liviuconcioiu: The problem is for people that might use this library... so we don't mess something.

liviuconcioiu: https://www.imei.info/phonedatabase/phones-myphone-pilipinas/

Simbiat: at the same time the imei.info has PH phone without "Pilipinas" https://www.imei.info/phonedatabase/myphone-mywx2-pro/
I still think adding PL and PH respectively to both brands would be the most neutral approach

sanchezzzhak (dismissed):

sanchezzzhak (approved):

PR #7773, opened 2024-08-11
Remove single quotes around brand names

liviuconcioiu:


regexes/device/mobiles.yml
```diff
@@ -33263,7 +33264,7 @@ Transpeed:
       model: '$1'
 
 # 8848 (everest 8848, 8848 titanium)
-'8848':
+8848:
```
sanchezzzhak, line 33267: return the quotes, since we get the numeric key

regexes/device/mobiles.yml
```diff
@@ -31975,7 +31976,7 @@ Owwo:
       model: 'Carnival B5'
 
 # 360 or names Qihoo 360, +360, QiKU 360
-'360':
+360:
```
sanchezzzhak, line 31979: return the quotes, since we get the numeric key

sanchezzzhak (approved):

PR #7783, opened 2024-08-16
Adds detection for Azupik, Everis, LW, Portfolio, Siswoo brands and improves detection for other brands

liviuconcioiu:


regexes/device/mobiles.yml
```diff
@@ -39843,6 +39877,12 @@ Everex:
   device: 'tablet'
   model: 'Fine 7B'
 
+# Everis (https://www.thewarehouse.co.nz/)
```
Simbiat, line 39880: Is this a hidden ad? 😅 I do not see how it's relevant to the brand

liviuconcioiu: > Is this a hidden ad? 😅 I do not see how it's relevant to the brand

Apparently, Everis is sold by The Warehouse, that's all I could find.

https://www.reddit.com/r/newzealand/comments/h0nnpw/everis_laptops_from_the_warehouse/

liviuconcioiu: _Everis" brand is exclusively marketed and distributed by
The Warehouse Group._

https://docs.thewarehouse.co.nz/red/pdfs/sports-outdoors/User%20manual_9401063611300_EVERIS%20SMART%20BAND_E4004_V3_approved%20on%2020th%20Aug.pdf

sanchezzzhak (approved):

PR #7807, opened 2024-09-03
Added support Sec-CH-UA-Form-Factors header for ClientHints

sanchezzzhak:
Specs: https://wicg.github.io/ua-client-hints/#sec-ch-ua-form-factors
form factor support for chrome 124+  https://chromestatus.com/feature/5162545698045952


sanchezzzhak:
maybe we should add examples to the documentation for client-hints?

For example, through JS, we can get them like this:

```js
function getHints() {
    return new Promise((resolve, reject) => {
      if (!navigator.userAgentData) {
        return resolve({})
      }
      let requestHints = [
        'brands',
        'mobile',
        'platform',
        'platformVersion',
        'architecture',
        'bitness',
        'wow64',
        'model',
        'formFactors',
        'uaFullVersion',
        'fullVersionList'];

      navigator.userAgentData.getHighEntropyValues(requestHints).then((result) => {
        resolve(JSON.parse(JSON.stringify(result)))
      }).catch(() => {
        resolve({});
      });
    })
  }

```

```js
getHints().then(data =>console.log(data)) 
```

sgiehl:
@sanchezzzhak Based on the definition formFactors is a `sf-list`. So it can contain multiple comma separated values. In theory something like `"Tablet","EInk"` would be a valid value. So we would actually need to handle that as array and not as string. Or did I misinterpret the definition 🤔

liviuconcioiu:
For reference, `Android` tablets are reporting as `Desktop`, but detection works.

tablets

sanchezzzhak:
@sgiehl yes there should be an array.

---

I think the form factor based on the user agent string, since there is no prefix (Table|Mobile|VR|TV|XR) is set by default to Desktop

Tests/ClientHintsTest.php
```diff
@@ -47,6 +47,7 @@ public function testHeadersHttp(): void
             'HTTP_SEC_CH_UA_PLATFORM'          => 'Ubuntu',
             'HTTP_SEC_CH_UA_PLATFORM_VERSION'  => '3.7',
             'HTTP_SEC_CH_UA_FULL_VERSION'      => '98.0.14335.105',
+            'HTTP_SEC-CH-UA-FORM-FACTORS'      => '"Desktop"',
```
liviuconcioiu, line 50:
```suggestion
            'HTTP_SEC_CH_UA_FORM_FACTORS'      => '"Desktop"',
```

sgiehl (approved):
Looks good to me now.

PR #7823, opened 2024-09-13
Skip recursive device parsing is UA client hints and the model name is empty

sanchezzzhak:
This improvement speeds up the work a little if the device model is not specified in clienthints, and the user agent is intentionally anonymized.

Full scan the list of regular expressions would be unnecessary.


liviuconcioiu:
```
2331 user agents in 11722 ms
2331 user agents in 3158 ms
```

Almost 4x speed improvement for those user agents.

liviuconcioiu (approved):

PR #7828, opened 2024-09-16

liviuconcioiu:


liviuconcioiu:
I think is best to split the Google bots into their own regex in the future, because Googlebot is to vague.

sanchezzzhak (approved):

PR #7843, opened 2024-09-21
Fix detecting device type by form factor of client hints

nommyde:
### Description:

In the method \DeviceDetector\Parser\Device\AbstractDeviceParser::parseClientHints,
condition `\array_key_exists($formFactor, $formFactors)` is always false because $formFactor is a string and $formFactors keys are integers ($formFactors is a list of strings). If we fix just this condition, then $deviceType gets the string value of self::getDeviceName, but it should be an int (we don't need the getDeviceName call)
Then, if we fix that, we see that 'deviceType' field is never used in the \DeviceDetector\Parser\Device\AbstractDeviceParser::parse method, so we need to fill $this->deviceType with it to get advantage of client hints when nothing else helps to detect device type.

I am not sure if the fixture I created fits your conventions, but it detects the described problems and does its job.

### DIFF
--- a/Parser/Device/AbstractDeviceParser.php
+++ b/Parser/Device/AbstractDeviceParser.php
@@ -2212,6 +2212,8 @@ abstract class AbstractDeviceParser extends AbstractParser
         }
 
         if (empty($matches)) {
+            $this->deviceType = $resultClientHint['deviceType'] ?? null;
+
             return $resultClientHint;
         }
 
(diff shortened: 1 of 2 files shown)


Parser/Device/AbstractDeviceParser.php
```diff
@@ -2297,13 +2299,11 @@ protected function parseClientHints(): ?array
...
-                        $deviceType = self::getDeviceName($deviceTypeId);
+            foreach ($formFactors as $formFactor) {
+                if (isset(self::$clientHintFormFactorsMapping[$formFactor])) {
+                    $deviceType = self::$clientHintFormFactorsMapping[$formFactor];
 
-                        break;
-                    }
+                    break;
                 }
             }
```
sanchezzzhak, line 2308:
```php
            foreach (self::$clientHintFormFactorsMapping as $formFactor => $deviceTypeId) {
                if (\in_array($formFactor, $formFactors)) {
                    $deviceType = $deviceTypeId;

                    break;
                }
            }
```

Nommyde: I get that we need to consider priority. But if the header contains `"EInk", "Watch"` or `"Desktop", "EInk"`, we'll recognize it as a tablet. Should we maybe drop the priority for eink?

sanchezzzhak: You can omit it, after desktop.

Nommyde: Thank you. I have updated the priorities and modified the fixture so that it will catch changing that algorithm to my previous version or the sorting of $clientHintFormFactorsMapping

sanchezzzhak (approved):

PR #7985, opened 2025-02-04
Adds detection for Pocket Internet Explorer browsers, smzdm, Zite apps and improves detection for Motorola Internet Browser, Safari, YouTube, YunOS

liviuconcioiu:


Parser/Client/Browser.php
```diff
@@ -90,6 +90,7 @@ class Browser extends AbstractClientParser
         'AH' => 'Aloha Browser Lite',
         'A8' => 'ALVA',
         '9A' => 'AltiBrowser',
+        'AY' => 'Aliyun Browser',
```
Simbiat, line 93: I do not think it's a browser. It is definitely some tool from [AliOS](https://en.wikipedia.org/wiki/AliOS), but I do not see it having some special browser. They do have OSS Browser, but that's not a web browser and I am not sure if it can browser internet, too. "AliyunOs-2012" seems to be the build of the firmware, which is, at least, partially confirmed by this [article](https://www.alizila.com/aliyun-cloud-based-smartphone-os-gets-its-first-major-upgrade/)

liviuconcioiu: All I could find was this:

https://github.com/WhichBrowser/Parser-PHP/blob/581d614d686bfbec3529ad60562a5213ac5d8d72/data/id-android.php#L20

And from `https://guopengliang.com/about/` - _Lead designer of Aliyun Browser v1.0 in GUI and UX design._

liviuconcioiu: Also:

https://github.com/ThaDafinser/UserAgentParserComparison/blob/04ae4e5a00394bc7ef7b894f9fffd20db3403de7/v5/user-agent-detail/0d/d4/0dd48269-21c0-48a6-b402-2cdde013bfe1.html#L21-L22

Parser/Client/Browser.php
```diff
@@ -410,6 +411,7 @@ class Browser extends AbstractClientParser
         'MS' => 'Mobile Silk',
         'MK' => 'Mogok Browser',
         'M8' => 'Motorola Internet Browser',
+        '4M' => 'Motorola WebKit',
```
Simbiat, line 414: I do not think this is a browser. Matomo already has MotorolaWebkit [here](https://github.com/matomo-org/device-detector/blob/cd2846d831b75f0a363701d634af3975731755eb/Tests/fixtures/smartphone-9.yml#L5091) and it's treated as Mobile Safari. But overall, I'd say it's most likely [this](https://github.com/matomo-org/device-detector/blob/cd2846d831b75f0a363701d634af3975731755eb/regexes/client/browsers.yml#L112). While Motorola did have own fork of WebKit, I just do not see them actually having a separate browser, it looks more like they called the _engine_ as _browser_.

Parser/Client/Browser.php
Simbiat: [this](https://github.com/ThaDafinser/UserAgentParserComparison/blob/04ae4e5a00394bc7ef7b894f9fffd20db3403de7/v5/user-agent-detail/0d/d4/0dd48269-21c0-48a6-b402-2cdde013bfe1.html#L41) confirms that this is related to AliyunOS 2.0. The resume says version 1.0  and it's possible that it did not even reach real devices, and not sure if it can be trusted without any links. Relying on header with `com.aliyun.mobile.browser` only also is not enough, IMHO. They might have had some mobile browser in development based on screenshots [here](https://www.techinasia.com/aliyun-alibaba-mobile-os), but these were pre-release, and possible that they just had custom icon and nothing else. It can also be earlier version of UC Browser, since  from quick search it seems to be the only actual browser owned by Alibaba

liviuconcioiu: I've removed it. For reference, UC Browser was acquired in 2014, two years later.

`https://ali-home.alibaba.com/en-US/about-alibaba-businesses-1747799156136607744`

sanchezzzhak (approved):

PR #8029, opened 2025-03-14
Add WeChat as known browser, to show it's icon

vibbow:
Related PR:
https://github.com/matomo-org/matomo-icons/pull/42

### Description:

### Review

### DIFF
--- a/Parser/Client/Browser.php
+++ b/Parser/Client/Browser.php
@@ -682,6 +682,7 @@ class Browser extends AbstractClientParser
         'W3' => 'Web Browser & Explorer',
         'W5' => 'Webian Shell',
         'W4' => 'WebDiscover',
+        'WC' => 'WeChat',
         'WE' => 'WebPositive',
         'W6' => 'Weltweitimnetz Browser',
         'WX' => 'Wexond',


Simbiat:
WeChat is not a browser, though, and it is present in Matomo as a mobile app, as it should be. Matomo also does not include icons, so unclear, what icon you want  to show through this change.

Simbiat:
Ah, you mean the `getBrowserLogo` in Matomo itself, not the DeviceDetector. And I am not sure if that code is the best way to approach things, but either way, forcing an application type does not seem like the way to go. Since the issue for this PR and for https://github.com/matomo-org/matomo/pull/23128 comes from plugin support, would it not make more sense to update the plugin instead? Or rather the icons in https://github.com/matomo-org/matomo-icons

sgiehl:
Hey there. @Simbiat is correct. WeChat is not a browser, but a mobile app. Matomo currently lacks the possibility to show icons for anything else than browsers. But that needs to be solved in Matomo rather than device detector. Therefor I'll close this PR.
Thanks anyway @vibbow for your approach!

Simbiat:
@sgiehl I was thinking of adding a small PHP class to my icons collection that would generate the link/path to icons, including fallback ones. I was then even thinking of adding support of it to DeviceDetector, so that if the class exists, it would add path to icon to respective entities. I wanted to do that after I complete the collection, but I can do that practically anytime anyway, dur to planned failback logic. If done, Matomo can be updated to rely on icons from DeviceDetector, if they are present, and my collection can be added as recommended. In my mind the only "blocker" is the "unofficial" status and the fact that there is an official icon pack. Personally, I would not mind  moving my collection to Matomo, and continue adding stuff like that and it could be used as "clients' icons" pack, while [matomo-icons](https://github.com/matomo-org/matomo-icons) is used for other icons. But I am open to discussion

PR #8096, opened 2025-09-14
Brands' name corrections, minor code adjustments

Simbiat:
### Description:

I recently finished collecting [icons](https://github.com/Simbiat/DeviceDetectorIcons) for all brands, and went through all my notes related to various potential inconsistencies that I've found while doing that. This PR includes implementation of all of those. Main focus is on brands' names:
1. Most are adjustments to case of letters used to match what I was able to find in official materials from the brand (manuals, official ads, logos, names on official website(s) (title, footer, "about us" sections, etc.)) or retailers' data when it was not available. If multiple variations were found, I used the most common. This is mostly for accuracy's sake.
2. For quite a few names were "expanded": either something like `Mobile` was added or abbreviation was spelled-out. Logic applied was same in as in point 1, but in some cases this was done to avoid name collision with brands with same name (or very similar one, for example `Suzuki`).
3. In a few rare cases brand was incorrect, usually model or model line name was used instead of a brand. Actual sub-brands were not to be touched, but with some obscure brands it might have happened.

This PR also corrected a couple of models and application names, as well as applied some recommendations from SAT in PHPStorm (native one and EA Ultimate), which should not have any negative affects, and in some cases may even slightly improve performance. The latter was done mostly because I got annoyed by IDE complaining about warnings, and wanted to fix at least some of them, where fix was obvious for me.

I do not have any strong opinion about most of the suggested changes (to brands or code), so if something needs to be reverted - that's not a problem. While this is being reviewed I will be working on icons for bots (thankfully there are way less of those), so I expect a similar PR after I am done with those, too.

### Review

### DIFF
--- a/Parser/OperatingSystem.php
+++ b/Parser/OperatingSystem.php
@@ -147,6 +147,7 @@ class OperatingSystem extends AbstractParser
         'OS2' => 'OS/2',
         'T64' => 'OSF1',
         'OBS' => 'OpenBSD',
+        'OHS' => 'OpenHarmony',
         'OVS' => 'OpenVMS',
         'OVZ' => 'OpenVZ',
         'OWR' => 'OpenWrt',
@@ -238,7 +239,7 @@ class OperatingSystem extends AbstractParser
         'Android'               => [
             'AND', 'CYN', 'FIR', 'REM', 'RZD', 'MLD', 'MCD', 'YNS', 'GRI', 'HAR',
             'ADR', 'CLR', 'BOS', 'REV', 'LEN', 'SIR', 'RRS', 'WER', 'PIC', 'ARM',
-            'HEL', 'BYI', 'RIS', 'PUF', 'LEA', 'MET',
+            'HEL', 'BYI', 'RIS', 'PUF', 'LEA', 'MET', 'OHS',
         ],
         'AmigaOS'               => ['AMG', 'MOR', 'ARO'],
         'BlackBerry'            => ['BLB', 'QNX'],
--- a/Tests/Parser/fixtures/oss.yml
+++ b/Tests/Parser/fixtures/oss.yml
@@ -6069,3 +6069,11 @@
     version: ""
     platform: ""
     family: Real-time OS
+-
+  user_agent: Mozilla/5.0 (Phone; OpenHarmony 5.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36 ArkWeb/4.1.6.1 Mobile MicroMessenger/8.0.9.39(0xf3100927) Weixin NetType/WIFI Language/zh_CN MMWEBID/2777 MMWEBSDK/202507060011 XWEB/1140307
+  os:
+    name: OpenHarmony
+    short_name: OHS
+    version: "5.1"
+    platform: ""
+    family: Android
--- a/regexes/oss.yml
+++ b/regexes/oss.yml
@@ -583,6 +583,10 @@
 ##########
 # Custom Android Roms
 ##########
+- regex: 'OpenHarmony(?:[/ ](\d+[.\d]+))?'
+  name: 'OpenHarmony'
+  version: '$1'
+
 - regex: 'HarmonyOS(?:[/ ](\d+[.\d]+))?'
   name: 'HarmonyOS'
   version: '$1'


Simbiat:
PHPCS and PHPStan errors are already fixed (needed to learn how to set them up, but no big deal). As for tests, I think I know how I broke them, or at least some of them. For part - missed some setup when renaming brands, but for some it looks like `in_array` fails due to use of `strict` mode now when working with numbers, because `$brand` can be an integer. Earlier code was probably "hiding" this and we might have been getting false positive/negatives. Anyway, I will address all those next week.

As for splitting... I do not mind, but considering that fixes for some tests will include corrections to the missed renaming here or there, I think it may confuse things even more. I probably should have setup PHPUnit on my side as well, but I thought it would be a brain-dead task 😅

Simbiat:
I was able to fix 101 tests. 40 - I am unsure yet how to fix, so any hints would be appreciated. I can group them:
1. 6 tests: `Anything LLM`, `OP.GG` and `VidJuice UniTube` apps fail due to the apps names being different from what is in the user-agent, if I understand correctly. I am unsure if it's possible to somehow "force" appropriate names instead of what's provided in the user-agent.
2. 12+1 tests: Smartisan OS is not recognized from user-agent itself, but it's still a separate build of Android. Because it's being recognized automatically need to somehow override OS name and Version based on brand and model. Unclear how to do that. This will automatically fix 1 extra test for OS tests coverage.
3. 21 tests: a bunch of device for `NEXON.UA`, `Mystery Electronics`, `X96mini`, `Vortex` are not recognized for some reason. The actual results are blanks, and can't figure out why.

Simbiat:
addressed most of the tests. 7 from the 3rd group remain. No clue yet why those fail. I suspect that they hit some other regex first, but not sure.

Also unsure why GitHub is not allowing me to resolve conflict, even when there is now technically no conflict 🤔 Or do I simply not have the rights to do so?

liviuconcioiu:
1. The problem is that you also moved the tv regex of `Vortex`. Also, if the user agent contains `Vestel` and is a tv, the regex should be but before it, otherwise will be detected as `Vestel`.

This should be kept in the `televisions.yml`:

```yml
# Vortex (sub-brand of https://www.altex.ro/)
Vortex:
  regex: 'VORTEX[;,)]'
  device: 'tv'
  models:
    - regex: '_TV_[A-Z0-9]+_([0-9]{4});'
      model: 'Smart TV ($1)'
```

And this in `mobiles.yml`:

```yml
# Vortex (sub-brand of https://www.altex.ro/)
Vortex:
  regex: '(V39TPHE01S|V32TPHDE1S|V24TPHDE1S)(?:[);/ ]|$)'
  device: 'tv'
  model: '$1'
```

Same, `NEXON.UA` should be put before `Vestel` in `televisions.yml`:

```yml
# NEXON.UA (https://nexon.ua)
NEXON.UA:
  regex: 'NEXON[;,)]'
  device: 'tv'
  models:
    - regex: '_TV_[A-Z0-9]+_([0-9]{4});'
      model: 'Smart TV ($1)'
```

and in `mobiles.yml`:

```yml
# NEXON.UA (https://nexon.ua)
NEXON.UA:
  regex: 'NEXON[ _](X[1235]\+|X[123589]|X10)(?:[);/ ]|$)'
  device: 'tv'
  model: '$1'
```

You have a missing (?:_UA)? here:

```yml
# Mystery Electronics
Mystery Electronics:
  regex: 'MYSTERY[_ ]TV[_ ](D\d{4}[A-Z]{2}\d{2})(?:_UA)?(?:[);/ ]|$)'
  device: 'tv'
  model: '$1'
```

For X96 mini you have a whitespace at the end in both regexes:

```yml
X96mini 
```

A brand could be put in both televisions.yml and mobiles.yml. The first handles most of the tvs that have generic user agents.

Also, most of the files, contain generic regexes, so any detection you add should be put before it. In the case of mobiles.yml, we have `Unknown` at the bottom.

2. To resolve the conflicts you have to merge `upstream/master` into `pr/8096`, then create a merge commit.



After, you will see a `Resolve conflicts before Merge` window.



You have to open each file and resolve the conflicts. Most of the time you have to click on `Accept Current Change` and save the file.



After, you have to continue with the merge



It can also be done directly from browser, but it becomes harder if you have a lot of conflicts. You can also use the command line if you don't have Github Desktop.

Simbiat:
I don't use GitHub Desktop, but I figured how to do this in browser, and... I just did not realize, that it was showing the editor, and not just the code in view mode, and that I needed to remove markers to activate the button. Now all is green. Guess I need to figure out a way to split the PR either through GitHub website or through PHPStorm's UI. Or git CLI, I guess. On the other hand, I am not in a rush to merge 🙃 Also even if I split just mobiles' or tvs' renames, for example, it will still be a big PR.

PR #8221, opened 2026-01-04

liviuconcioiu:


regexes/bots.yml
```diff
@@ -2189,14 +2214,30 @@
...
   name: 'Yandex Bot'
   category: 'Search bot'
   url: 'https://yandex.com/support/webmaster/robot-workings/check-yandex-robots.html'
   producer:
     name: 'Yandex LLC'
     url: 'https://yandex.com/company/'
 
+- regex: 'YandexAdditionalBot'
+  name: 'YandexAdditionalBot'
+  category: 'AI Assistant'
```
sgiehl: According to their documentation that useragent should only be used to fetch the robots.txt in order to determine if the data (already) indexed by the normal search crawler can be used for AI. It doesn't perform indexing requests itself.
```suggestion
  category: 'AI Search Crawler'
```

regexes/bots.yml
```diff
@@ -2189,14 +2214,30 @@
...
+  url: 'https://yandex.com/support/webmaster/robot-workings/check-yandex-robots.html'
+  producer:
+    name: 'Yandex LLC'
+    url: 'https://yandex.com/company/'
+
+- regex: 'YandexAdditional'
```
sgiehl, line 2225: This one isn't listed in their documentations. Maybe we can simply group it with the previous one - if it actually exists

sgiehl (changes requested):

regexes/bots.yml
liviuconcioiu: It's at the bottom, here: `https://yandex.com/support/webmaster/en/robot-workings/check-yandex-robots.html`. It is the same bot, but different token. I merged them.

yandex

sgiehl (approved):

PR #8288, opened 2026-05-28

### LINKED ISSUE #8307: Sentione bot uncorrectly classified as googlebot
I've noticed that [bots.yml](https://github.com/matomo-org/device-detector/blob/master/regexes/bots.yml) regexes miss one for detecting sentione web crawler user agent, which is 

```SentiBot www.sentibot.eu (compatible with Googlebot)```

More informations about this bot are available here: [SentiBot.eu](https://sites.google.com/senti1.com/sentibot-eu/home)

### PR DESCRIPTION
**Pull request by liviuconcioiu**

### CODE REVIEW

File: regexes/bots.yml [hunk 1]
```diff
@@ -5857,14 +5865,26 @@
...
+- regex: '(CheckerProxy|CMSLookup|CSE-Scanner|DomainProbe|FreePBX-Scanner|libredtail-http|SecurityScanner|Shodan-Pull)'
```
Simbiat: I am not sure libredtail-http should be treated a security checker, since from what I read it's a a malicious crypto thing

regexes/bots.yml
```diff
@@ -5929,16 +5937,38 @@
     name: 'MARKETGOO INTERNET, S.L.'
     url: 'https://www.marketgoo.com/'
 
+- regex: 'RootEvidence'
+  name: 'RootEvidence'
```
Simbiat, line 5949: Shouldn't this have a space, since the company's name has it?

regexes/bots.yml
```diff
@@ -5929,16 +5937,38 @@
...
+
+- regex: 'TrustBot'
+  name: 'TrustBot'
+  category: 'Crawler'
+  url: 'https://truster.info/en/'
+
+- regex: '(recon-(?:crawler|engine))'
+  name: '$1'
+  category: 'Security Checker'
+  url: 'https://ringsix.com/'
```
Simbiat: Are you sure that's the one? There is just an mp4 on the website. Searching `ringsix` does not find anything security related, too. Nothing obvious for `recon crawler` and `recon engine`. There is https://github.com/samsepi0x0/ReconCrawler but I do not see it setting any user-agent.

Simbiat: Ah, I see that domain used in one of the agents, but considering that others are used as well, I do not think this can be trusted as the source of the bot. Also not sure if this can be treated as a security checker as well then.

Tests/fixtures/bots.yml
```diff
@@ -10340,3 +10340,89 @@
...
+    url: https://geedo.com/product-finder/
+-
+  user_agent: Mozilla/5.0 (compatible; NetScope-DomainScanner/1.0; security research scanner)
+  bot:
+    name: NetScope-DomainScanner
+    category: Security Checker
+-
+  user_agent: Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; SleepBot/1.0; +http://sleepbot.com/) Chrome/131.0.0.0 Safari/537.36
+  bot:
+    name: SleepBot
```
Simbiat, line 10389: The site say `Sleepbot`, not `SleepBot`

Tests/fixtures/bots.yml
```diff
@@ -10340,3 +10340,89 @@
...
+-
+  user_agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko; GeedoShopProductFinder) Chrome/142.0.0.0 Safari/537.36
+  bot:
+    name: GeedoShopProductFinder
+    category: Crawler
+    url: https://geedo.com/product-finder/
+-
+  user_agent: Mozilla/5.0 (compatible; NetScope-DomainScanner/1.0; security research scanner)
+  bot:
+    name: NetScope-DomainScanner
```
Simbiat, line 10384: Possibly an app https://xdaforums.com/t/beta-test-netscope-network-tools-app.4759000/ and not a bot. They do not have it publicly available, though, but description on XDA is the closest to a domain scanner. https://www.netskope.com could be it, too, but I doubt they would make an intentional mistype.

Tests/fixtures/bots.yml
liviuconcioiu: I know, but we should add bot names exactly like the token, where possible. All websites or robots.txt don't rename them.

Tests/fixtures/bots.yml
liviuconcioiu: Yeah, I don't think is that.

regexes/bots.yml
liviuconcioiu: I know, but we should add bot names exactly like the token, where possible. All websites or robots.txt don't rename them.

regexes/bots.yml
liviuconcioiu: I made it a generic bot.

sanchezzzhak (dismissed):

sanchezzzhak (approved):

PR #8298, opened 2026-06-07
Adds detection for Vega OS

liviuconcioiu:
`Kepler` is the internal OS name for `Vega OS`.

```
https://developer.amazon.com/apps-and-games/vega
https://community.amazondeveloper.com/t/vega-ids-uuid-ad-id-vendor-tracking-and-device-info/11447
https://developer.amazon.com/docs/fire-tv/fire-os-overview.html
```


sanchezzzhak (approved):

sanchezzzhak:
https://developer.amazon.com/docs/device-specs/identify-fire-tv-devices.html
there is already a new OS on this list

PR #8332, opened 2026-07-24
Adds detection for Lightpanda and Improves detection for UC Browser

liviuconcioiu:


sanchezzzhak (approved):

cnkk:
@liviuconcioiu @sanchezzzhak shouldn't Lightpanda not rather be classified as bot instead?

liviuconcioiu:
No, because is a headless browser, same like `Headless Chrome`, `Headless Edge`. @sanchezzzhak?

liviuconcioiu:
Also, libraries like `curl` which is used 99% for scraping is detect as library not bot.

PR #8370, opened 2026-08-18

liviuconcioiu:


sanchezzzhak (dismissed):

Tests/fixtures/bots.yml
```diff
@@ -10638,3 +10638,44 @@
     name: WebatlaBot
     category: Crawler
     url: https://bot.webatla.com/
+-
+  user_agent: MyCellar/1.0 (https://www.mycellar.world)
+  bot:
+    name: Generic Bot
```
Simbiat: Why not `MyCellar`?

liviuconcioiu: I changed it. Some might use link in user-agent as a spam method to increase SEO visibility.

sanchezzzhak (approved):

PR #8379, opened 2026-09-09
Adds brand detection for eight webOS televisions

Amoifr:
Continues #8012, alongside #8378 which covers the LG entry.

All the remaining user agents in the list already resolve to `type: tv` with the right model, only the brand is missing. Six of them belong to brands that already exist in `mobiles.yml`, so this adds the models to those entries and nothing else. The brands that would be new are left out on purpose, they need a short code of their own and live in #8380.

Every model below was checked against a source outside this repository before being kept.

| Brand | Model | Source | Note |
| --- | --- | --- | --- |
| GEEPAS | GLED7580SWUHD | uae.geepas.com, jomla.ae | had no `models` list yet |
| JVC | LT-42NQ5155 | jvctv.com.au, bigw.com.au | added to the existing alternation |
| Hyundai | HYLED5017W4KM | olimpica.com, exito.com (Colombia) | brand defaults to `smartphone`, so the model carries `device: 'tv'` |
| iKon | IK-58A71WOS | Lulu KSA, shops.ae, ikonhome.com | brand defaults to `tablet`, same |
| Viendo | V65D2 | amazon.sa | `(V65G2)` widened to `(V65[DG]2)` |
| WANSA | WUD65KWO60 | xcite.com, WANSA being Xcite's own brand | added to the existing alternation |

Two entries were dropped during review, both because the string does not name a product:

* `Sunny 14AT542` is a mainboard reference, listed by parts sellers as `TPD.RT2864.PB802 (T) 14AT542` on a set whose actual model is `SN43DAL540-0276`.
* `Vega E65A71B-S`: `E65A71B` is a white-label chassis reference sold by Voltic in the Maldives and, in its 55" and 50" siblings, by Webor in Nepal. Nothing ties it to Vega.

#### Testing

The full suite passes: 40029 tests, 186673 assertions. I also ran it once with the regex changes but without the new fixtures, to check that none of the new patterns steals a device from an existing fixture, and it stayed green. The Ruby validator from the `Validate regular Expressions` workflow was run locally as well.

Heads-up: this touches the same fixture file as #8378, so whichever lands second may need a trivial rebase.

Issue #8012, opened 2025-02-25
WebOS TVs


EfekanPekel:
- [ ] Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chr0me/87.0.4280.88 Safari/537.36 LG Browser/8.00.00(LGE; ATV75UHDW-0324; 04.42.26; 0x00000001; DTV_C22P); webOS.TV-2022; LG NetCast.TV-2013 Compatible (LGE, ATV75UHDW-0324, wireless)
Brand: Bauhn (https://bauhn.com.au/75-4k-ultra-hd-webos-smart-tv/)
- [ ] Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chr0me/87.0.4280.88 Safari/537.36 LG Browser/8.00.00(LGE; GA50F-OS; 03.34.35; 0x00000001; DTV_C22P); webOS.TV-2022; LG NetCast.TV-2013 Compatible (LGE, GA50F-OS, wireless)
Brand: Gazal (https://jo.opensooq.com/en/search/259394915)
- [ ] Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chr0me/87.0.4280.88 Safari/537.36 LG Browser/8.00.00(LGE; GLED7580SWUHD; 04.42.26; 0x00000001; DTV_C22P); webOS.TV-2022; LG NetCast.TV-2013 Compatible (LGE, GLED7580SWUHD, wireless)
Brand: Geepas (https://www.vinodpatel.com.fj/products/geepas-tv-75-4k-uhd-smart-webos-tv)
- [ ] Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chr0me/94.0.4606.128 Safari/537.36 LG Browser/8.00.00(LGE; LT-42NQ5155; 03.11.77; 0x00000001; DTV_C23L); webOS.TV-2022; LG NetCast.TV-2013 Compatible (LGE, LT-42NQ5155, wireless)
Brand: JVC (https://www.hificorp.co.za/jvc-42-inch-qled-smart-tv-lt42nq5155)
[...]

PR #8380, opened 2026-09-09
Adds eight webOS television brands to the mobile parser

Amoifr:
Finishes #8012, after #8378 (LG) and #8379 (the eight brands that already had a rule in `mobiles.yml`).

These eight are the ones the mobile parser did not know at all. As before, the user agents already resolve to `type: tv` with the right model, only the brand was missing.

Three of them are not new brands at all, which is worth knowing: `Linsar`, `Ssmart` and `TOKYO` already have a short code, but their only rule lives in `televisions.yml`, which the television parser uses for HbbTV user agents. A webOS user agent goes through the mobile parser instead, so it never saw them. They get a rule in `mobiles.yml` and keep their existing code.

`TOKYO` is spelled in capitals on purpose, that is how the brand is registered (`'Y3' => 'TOKYO'`), and the key has to match for the brand to resolve.

The five genuinely new brands:

| Code | Brand | Model |
| --- | --- | --- |
| HMZ | Homez | HZ50UHDWOS |
| NDC | Nadco | NC-65T4K-SMART |
| STK | Star Track | ST50K-LG2200WEB |
| TZN | Telezone | TZ50WF1 |
| VON | Von | VEL55USVW |

#### Testing

Full suite green: 43732 tests, 199387 assertions, `testDuplicateBrands` included. The Ruby validator from the `Validate regular Expressions` workflow was run locally as well.

Heads-up: like #8379, this appends to `Tests/fixtures/tv-5.yml`, so whichever lands second may need a trivial rebase.


regexes/device/mobiles.yml
```diff
@@ -45564,6 +45564,54 @@ Galatec:
...
+  device: 'tv'
+  model: '$1'
+
+# Nadco
+Nadco:
+  regex: '(NC-65T4K-SMART)(?:[);/ ]|$)'
+  device: 'tv'
+  model: '$1'
+
+# Ssmart
```
Simbiat: https://ssmart.uz

regexes/device/mobiles.yml
```diff
@@ -45564,6 +45564,54 @@ Galatec:
...
+  device: 'tv'
+  model: '$1'
+
+# Linsar (linsar.com)
+Linsar:
+  regex: '(LS75UHDNF)(?:[);/ ]|$)'
+  device: 'tv'
+  model: '$1'
+
+# Nadco
```
Simbiat: https://nadcostore.com https://www.facebook.com/nadco.official

regexes/device/mobiles.yml
Amoifr: Thank you, applied. The model checks out there too: idea.uz sells it as "Телевизор Ssmart FSV22 43 WEB OS", which is the `43FSV22` from the user agent. I should have found this myself instead of leaving it to you.

regexes/device/mobiles.yml
Amoifr: Thank you, applied both. `nadcostore.com` does not resolve from here at the moment, so I kept your Facebook link next to it rather than shipping a single reference I cannot reach.

Worth noting for whoever searches later: `nadco.com.sa` is a Saudi kitchen manufacturer, a different company entirely.

Amoifr:
@Simbiat I owe you an apology. This PR went out without the research it needed: the brands were read off the model prefixes in the user agents (`HZ...` to Homez, `TZ...` to Telezone) rather than checked against the manufacturers. That is guessing dressed up as identification, and you should not have had to go and find those two links for me.

I have now checked all eight against sources outside this repository, and pushed the corrections in efb419e.

The eight attributions themselves do hold up, but one reference was wrong and three were missing:

| Model | Brand | Reference | |
| --- | --- | --- | --- |
| HZ50UHDWOS | Homez | none found | sold by eXtra and noon in KSA/UAE |
| LS75UHDNF | Linsar | `linsar.com.au` | **was wrong**: `linsar.com` is the UK Freeview line, the `LSxxUHDNF` family is the Australian webOS range |
| NC-65T4K-SMART | Nadco | `nadcostore.com` + your Facebook link | added |
| 43FSV22 | Ssmart | `ssmart.uz` | added |
| ST50K-LG2200WEB | Star Track | `startrackworld.com` | already correct |
| TZ50WF1 | Telezone | none found | sold by Saudi resellers, confirmed webOS |
| TVTOKHF32WB | TOKYO | `tokyo.com.py` | added, Tokyo Electrodomésticos in Paraguay |
| VEL55USVW | Von | `von.co.ke` | already correct, the manufacturer's own catalogue page |

Homez and Telezone have no manufacturer site I can find, so their comment stays bare rather than carrying one I made up. Two traps the search did avoid, in case they save someone else time: `homez.com` is an unrelated furnished-rental platform, and `nadco.com.sa` is a Saudi kitchen manufacturer that happens to share the name.

One question while you are here. Ssmart already carries a reference in `televisions.yml`, `https://ssmartelectronics.ru/`, and `ssmart.uz` is Uzbek. Same company with two regional sites, or two brands sharing a name? If it is the latter they should probably not share the `SSM` short code, and I will split them.

Full suite still green after the change: 40031 tests, 186741 assertions.

Simbiat:
That's why AI should be targeted to specific actions... I did not go out looking for websites, I am searching for logos, and I just coincidentally found websites. ssmartelectronics.ru and ssmart.uz are the same company.

sanchezzzhak (dismissed):

Tests/fixtures/tv-5.yml
```diff
@@ -6000,4 +6000,146 @@
     brand: Apple
     model: Apple TV 3 4K
   os_family: iOS
+  user_agent: Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chr0me/68.0.3440.106 Safari/537.36 LG Browser/8.00.00(LGE; HZ50UHDWOS; 04.50.63; 0x00000001; DTV_C20P); webOS.TV-2020; LG NetCast.TV-2013 Compatible (LGE, HZ50UHDWOS, wireless)
```
sanchezzzhak, line 6005: missing before section test `-`

Amoifr: Good catch, thank you. The rebase had eaten the `-` and also the `browser_family: Unknown` line that ended the entry above, so the two fixtures had merged into one and the suite stayed green while silently losing a test case. I rebuilt the file as master plus my own entries: 344 entries now, and the suite reports one test more than before.

sanchezzzhak (approved):

PR #8382, opened 2026-09-11
Adds detection for various brands and improves detection for existing ones

liviuconcioiu:
```
flow - https://upload.wikimedia.org/wikipedia/commons/thumb/b/be/Flow_logo_2021.svg/250px-Flow_logo_2021.svg.png
```


Tests/fixtures/car_browser.yml
```diff
@@ -1079,3 +1079,25 @@
     model: DUDU7
   os_family: Android
   browser_family: Unknown
+-
+  user_agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36
+  headers:
+    Sec-CH-UA: '"Chromium";v="148.0.7778.167", "Google Chrome";v="148.0.7778.167", "Not/A)Brand";v="99.0.0.0"'
+    Sec-CH-UA-Platform-Version: 12.0.0
+    Sec-CH-UA-Model: uis8581a2h10_Automotive
```
sanchezzzhak, line 1087: This is not a specific brand of the device (like Teyes, Mekede or Kingbeats), but the technical system name of the motherboard and processor, which is embedded in the operating system of your Android screen

it is logical to leave only the device type
```yml
 type: car browser
 brand: ""
 model: ""
```

liviuconcioiu: Thanks! Can't set a empty model, since we fallback to the client hints model. Probably we need to change the code in the future for cases like this.

sanchezzzhak (approved):

PR #8385, opened 2026-09-14
Improves mobile detection from client hints

dmitryuk:
* Makes the `mobile` property in `ClientHints` nullable to differentiate between absent headers and an explicit false value
* Updates `DeviceDetector::isMobile()` to return the exact boolean value if provided by client hints, rather than only checking for true
* Updates related assertions in `ClientHintsTest` to expect null when the header is not present

### Description

### Checklist

- [✔/✖/NA] I have understood, reviewed, and tested all AI outputs before use
- [✔/✖/NA] All AI instructions respect security, IP, and privacy rules

### Review

- [x] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
- [x] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behaviour of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
- [x] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
- [x] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
- [x] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
- [x] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
- [x] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
- [x] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
- [x] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
- [x] [New documentation added or existing documentation updated](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)


DeviceDetector.php
```diff
@@ -379,9 +379,9 @@ public function isTouchEnabled(): bool
      */
     public function isMobile(): bool
     {
-        // Client hints indicate a mobile device
-        if ($this->clientHints instanceof ClientHints && $this->clientHints->isMobile()) {
-            return true;
+        // Client hints can indicate a mobile device
+        if ($this->clientHints instanceof ClientHints && null !== $this->clientHints->isMobile()) {
```
sanchezzzhak, line 383: && null !== $this->clientHints->mobile

ClientHints.php
```diff
@@ -130,11 +131,11 @@ public function __get(string $variable)
...
 
     /**
-     * Returns if the client hints
+     * Returns bool if detected by the client hints
      *
-     * @return bool
+     * @return bool|null
      */
-    public function isMobile(): bool
+    public function isMobile(): ?bool
```
sanchezzzhak, line 138: Can, is, has methods must return a bool type only

ClientHints.php
```diff
@@ -130,11 +131,11 @@ public function __get(string $variable)
...
     {
         return $this->mobile;
```
sanchezzzhak, line 140: return (bool) $this->mobile

sanchezzzhak (changes requested):

ClientHints.php
dmitryuk: Can you assist how to call the method? Since I need 3 states, not only `true` or `false` values:
- mobile detected
- not mobile detected
- not detected since no client hints headers provided by request

sanchezzzhak: You can access this property $clienthints->mobile to get three states.
https://github.com/matomo-org/device-detector/pull/8385/changes#diff-92cd8f8f62a550e5f2ed4718c04c3d21da4c039f9f96ddf6d20a4626d22fc359L123

See magic method __get for current class

If we make the changes as I proposed, we won’t break compatibility with clienthits for those who use it directly.

dmitryuk: all notes fixed, take a look

PR #8387, opened 2026-09-16
Improves detection for Apple and macOS

liviuconcioiu:


sanchezzzhak (approved):

sanchezzzhak:
Update convention naming 
https://github.com/matomo-org/device-detector/wiki/Convention-naming-device-models

add
⚠️ **Important note for the `mini` / `Mini` modifier:**
In the vast majority of cases, the token must be capitalized as **`Mini`**. The use of the completely lowercase **`mini`** is permitted **exclusively** when the manufacturer has officially designated it as an integral part of the trademark in lowercase at the brand level (e.g., in marketing materials and official device specifications). In all other ambiguous situations, the capitalized version should be used by default.

PR #8389, opened 2026-09-17
Adds detection for YaOS and YodaOS-Master

liviuconcioiu:


regexes/oss.yml
```diff
@@ -5,6 +5,20 @@
...
 
+##########
+# YodaOS-Master (https://github.com/yodaos-project/yodaos | https://global.rokid.com/blogs/articles/what-is-rokid-yodaos-master-how-does-it-enhance-ar-immersive-experience)
+##########
+- regex: 'RES102(?:[);/ ]|$)'
+  name: 'YodaOS-Master'
+  version: ''
+
+##########
+# YaOS (https://alice.yandex.ru/yaos)
```
sanchezzzhak, line 16: YaOS is actively used in RAZZ brand TVs.

sanchezzzhak (approved):

regexes/oss.yml
liviuconcioiu: Yeah, I thought of doing something like all Yandex browsers on TV, but then it may also run on Android, unless we know for sure that all devices for a brand run it.

sanchezzzhak: all browsers on Yandex cannot be with the YaOS OS, for example, I have a Sony bravia at home where yandex browser is installed by default.  There's an old LG at work, and Yandex is there by default too.

so we don't have to bind to the browser.

